PR #17, opened 2019-08-14
BufferCC misuse remove 

jijingg:


source/SpinalHDL/Design errors/clock_crossing_violation.rst
```diff
@@ -82,22 +82,5 @@ You can also specify that two clock domains are syncronous together.
...
-
-     val area_clkB = new ClockingArea(clkB){
-       val syncA = BufferCC(regA, U(0))
-
-       val tmp = syncA + syncA
-       regB := tmp
-     }
-   }
+Signal Bits or Gray-coded Bits can BufferCC to cross different clockDomain, 
+But BufferCC can't be used for general multi-Bits cross-domain process !
```
wifasoi: Maybe this is a bit extreme...
I think the best course of action is to change the  example code to something like "external button/IO"

and use an admonition block like `.. note::` or `.. warning::`

source/SpinalHDL/Structuring/clock_domain.rst
```diff
@@ -373,28 +373,6 @@ SpinalHDL checks at compile time that there is no unwanted/unspecified cross clo
...
-       val reg = RegNext(io.dataIn) init(False)
-     }
-
-     // BufferCC to avoid metastability issues
-     val area_clkB = new ClockingArea(clkB){  
-       val buf1   = BufferCC(area_clkA.reg, False)
-     }
-
-     io.dataOut := area_clkB.buf1
-   }
```
wifasoi: like above, maybe just change the example and add some admonition block

Dolu1990:
I think enumerating the usage of BufferCC would be great, as discused on gitter : 
- cross clock boolean value
- cross clock gray counter value
- protecting the output clock domain logic from metastable transiants, knowing that it can read transitional values. (but at least well etablished)

source/SpinalHDL/Structuring/clock_domain.rst
jijingg: yes, i agree ,this example code no problem.  just modify the top description is ok.

jijingg:
hi @wifasoi @Dolu1990 , I made a change  according to the comments , but not found re-request reviews buttons on my pages. what should i do next ?

source/SpinalHDL/Structuring/clock_domain.rst
```diff
@@ -373,7 +373,8 @@ SpinalHDL checks at compile time that there is no unwanted/unspecified cross clo
      io.dataOut := area_clkB.buf1
    }
 
-An even simpler method is to use the cross clock domain buffer ``BufferCC(input: T, init: T = null, bufferDepth: Int = 2)`` provided by ``spinal.lib._`` to prevent metastability
+In general, just use 2 filp-flops with dest-clock to prevent metastability. you can use  ``BufferCC(input: T, init: T = null, bufferDepth: Int = 2)`` provided by ``spinal.lib._`` instead.
```
wifasoi: mmm, this phrase is a bit confusing. What about something like this:

> In general you can use 2 or more flip-flop driven by the destination clock domain to prevent metastability. The ``BufferCC(input: T, init: T = null, bufferDepth: Int = 2)`` function provided in ``spinal.lib._`` will instantiate the necessary flip-flop (the number of flip flop will depends on the ``bufferDepth``) to mitigate the phenomena.

feel free to change my example frase (like maybe "the phenomena" can be changed in "metastability problem" (my native language is frown upon you if you repeat a word in the same sentence, or in the same chapter.. but in English is ok) I leave it to your judgment.

wifasoi:
> hi @wifasoi @Dolu1990 , I made a change according to the comments , but not found re-request reviews buttons on my pages. what should i do next ?

Sorry, it was my mistake.. i don't use the github code review often an I messed up :)

source/SpinalHDL/Structuring/clock_domain.rst
jijingg: ok，your suggestion is  even better.  i will merge it soon.

wifasoi:
Checked and merged to master.
Thank you for you contribution.

PR #18, opened 2019-10-09
Simple install verilator

MindShow:
Simple install verilator

### DIFF
--- a/source/SpinalHDL/Simulation/install.rst
+++ b/source/SpinalHDL/Simulation/install.rst
@@ -64,6 +64,19 @@ From a fresh install of MSYS2 MinGW 64-bits, you will have to run the following 
                       git flex\
                       mingw-w64-x86_64-cmake
 
+   pacman -S mingw-w64-x86_64-verilator
+   
+   #Add C:\msys64\usr\bin;C:\msys64\mingw64\bin to you windows PATH
+
+.. code-block:: sh
+
+   pacman -Syuu
+   #Close the MSYS2 shell once you're asked to
+   pacman -Syuu
+   pacman -S --needed base-devel mingw-w64-x86_64-toolchain \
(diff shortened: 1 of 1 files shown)


source/SpinalHDL/Simulation/install.rst
```diff
@@ -55,6 +55,19 @@ Then you should be able to run SpinalSim + verilator from your Scala project wit
 
 From a fresh install of MSYS2 MinGW 64-bits, you will have to run the following commands inside the MSYS2 MinGW 64-bits shell (enter commands one by one):
```
wifasoi: I think a subparagraph is necessary to distinguish the step.. 
```suggestion

from the minGW packet manager
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

source/SpinalHDL/Simulation/install.rst
```diff
@@ -55,6 +55,19 @@ Then you should be able to run SpinalSim + verilator from your Scala project wit
...
+   #Close the MSYS2 shell once you're asked to
+   pacman -Syuu
+   pacman -S --needed base-devel mingw-w64-x86_64-toolchain \
+                      git flex\
+                      mingw-w64-x86_64-cmake
+
+   pacman -S mingw-w64-x86_64-verilator
+   
+   #Add C:\msys64\usr\bin;C:\msys64\mingw64\bin to you windows PATH
+
```
wifasoi: I think a subparagraph is necessary to distinguish the step.. 
```suggestion

from source
~~~~~~~~~~~

```

source/SpinalHDL/Simulation/install.rst
MindShow: I can't agree with you more

PR #32, opened 2020-02-16
Fix small errors in Getting Started

andreasWallner:
Repo will checkout into folder "SpinalTemplateSbt"
"sbt run" will open a menu, added a comment so that the docs do not seem out of date


Dolu1990:
Thanks :)

PR #34, opened 2020-04-17
Add StreamJoin + cleanup of StreamFork

piessteady:
Small cleanup of Stream documentation.

### DIFF
--- a/source/SpinalHDL/Libraries/stream.rst
+++ b/source/SpinalHDL/Libraries/stream.rst
@@ -416,19 +416,39 @@ When you have multiple Streams and you want to arbitrate them to drive a single 
    * - onArgs(inputs : Stream[T]*)
      - Stream[T]
 
+StreamJoin
+^^^^^^^^^^
+
+This utile takes multiple input streams and wait until all of them fire before letting all of them through.
+
+.. code-block:: scala
+
+   val cmdJoin = Stream(Cmd())
+   cmdJoin.arbitrationFrom(StreamJoin.arg(cmdABuffer, cmdBBuffer))
+
 
 StreamFork
 ^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


source/SpinalHDL/Libraries/stream.rst
```diff
@@ -425,10 +435,8 @@ This utile take its input stream and duplicate it outputCount times.
 .. code-block:: scala
 
    val inputStream = Stream(Bits(8 bits))
-   val dispatchedStreams = StreamDispatcherSequencial(
-     input = inputStream,
-     outputCount = 3
-   )
+   val (outputstream1, outputstream2) = StreamFork(inputStream,portCount=2)
```
Dolu1990, line 438: This seem broken to me, as StreamFork return a Vec[Stream[T]],
Shouldn't it be either : 

```scala
    val inputStream = Stream(Bool)
    val outputStreams = StreamFork(inputStream,portCount=2)
    // You get outputStreams(0) and outputStreams(1)
```

or 

```scala
    val inputStream = Stream(Bool)
    val (outputstream1, outputstream2)= StreamFork2(inputStream)
```

?

Dolu1990:
Hi ^^
Thanks for the changes :), just one point above about the stream fork seem weird to me, what was the intend ?

source/SpinalHDL/Libraries/stream.rst
piessteady: Yes, wrongly changed from StreamFork2 to StreamFork without checking the return type would be different. Will change it.

piessteady:
Changed and updated the documentation to the comment from the code. Is a better explanation.

Dolu1990:
Thanks :)

PR #37, opened 2020-05-06
Add detailed protocol description to Stream

andreasWallner:
- Add detailed description of expected behavior of the signals in a
  Stream bundle
- Fix minor typos

### DIFF
--- a/source/SpinalHDL/Libraries/stream.rst
+++ b/source/SpinalHDL/Libraries/stream.rst
@@ -74,6 +74,17 @@ There is some examples of usage in SpinalHDL :
 * An UART controller could directly use the write port to drive UART pins and only consume the transaction at the end of the transmission.
   Be careful with that.
 
+Semantics
+---------
+
+When manually reading/driving the signals of a Stream keep in mind that:
+
+* After being asserted, ``valid`` may only be deasserted once the current payload was acknowleged. This means ``valid`` can only toggle to 0 the cycle after a the slave did a read by asserting ``ready``.
+* In contrast to that ``ready`` may change at any time. 
+* A transfer is only done on cycles where both ``valid`` and ``ready`` are asserted.
+* ``valid`` of a Stream must not depend on ``ready`` in a combinatorial way and any path between the two must be registered.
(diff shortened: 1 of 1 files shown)


source/SpinalHDL/Libraries/stream.rst
```diff
@@ -299,8 +309,8 @@ You can instanciate the dual clock domain version of the fifo by the following w
 StreamCCByToggle
 ^^^^^^^^^^^^^^^^
 
-| Component that provide a Stream cross clock domain bridge based on toggling signals.
-| This way of doing cross clock domain bridge is characterized by a small area usage but also a low bandwidth.
+| Component that provides connectes Streams across clock domains based on toggling signals.
```
wifasoi, line 312: mmm, I think it should be:
```suggestion
| Component that connects Streams across clock domains based on toggling signals.
```

source/SpinalHDL/Libraries/stream.rst
```diff
@@ -74,6 +74,16 @@ There is some examples of usage in SpinalHDL :
...
 
+Semantics
+---------
+
+When manually reading/driving the signals of a Stream keep in mind that:
+
+* After being asserted, ``valid`` may only be deasserted once the current payload was acknowleged. This means ``valid`` can only toggle to 0 the cycle after a the slave did a read by asserting ``ready``.
+* In contrast to that ``ready`` may change at any time. 
+* A transfer is only done on cycles where both ``valid`` and ``ready`` are asserted.
+* ``valid`` of a Stream must not combinatorically depend on ``ready`` and must be registered.
```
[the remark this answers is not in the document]
andreasWallner: You are right, it would actually be combinatorially ... which does not make it much better though.
After thinking about it: since ready may anyhow change it any time it does kinda not make sense to depend on it at all... Therefore I added that as a recommendation.

source/SpinalHDL/Libraries/stream.rst
andreasWallner: agreed

wifasoi:
Thank you for contributing on the "GODDENING" of the documentation :D

PR #43, opened 2020-08-26
Copyediting fixes for Structuring section.

philipaconrad:
This PR includes spelling/grammar/formatting fixes, as well as the
following:

 - `addPrePostTask` on the blackboxing page is now updated to be
   `addPrePopTask`, along with some explanatory text.

### DIFF
--- a/source/SpinalHDL/Structuring/clock_domain.rst
+++ b/source/SpinalHDL/Structuring/clock_domain.rst
@@ -9,7 +9,7 @@ Clock domains
 Introduction
 ------------
 
-In *SpinalHDL*\ , clock and reset signals can be combined to create a **clock domain**. Clock domains can be applied to some areas of the design and then all synchronous elements instantiated into those areas will then **implicitly** use this clock domain.
+In SpinalHDL, clock and reset signals can be combined to create a **clock domain**. Clock domains can be applied to some areas of the design and then all synchronous elements instantiated into those areas will then `implicitly` use this clock domain.
 
 Clock domain application works like a stack, which means that if you are in a given clock domain you can still apply another clock domain locally.
 
@@ -44,7 +44,7 @@ This definition takes five parameters:
      - Clock signal that defines the domain
      - 
    * - ``reset``
-     - Reset signal. If a register which needs a reset and the clock domain doesn't provide one, an error message will be displayed
+     - Reset signal. If a register exists which needs a reset and the clock domain doesn't provide one, an error message will be displayed
      - null
    * - ``softReset``
      - Reset which infers an additional synchronous reset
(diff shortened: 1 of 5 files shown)


philipaconrad:
The force-push was to correct my commit message. I wound up typing `addPrePostTask` twice in the commit message, which was a bit silly.

source/SpinalHDL/Structuring/clock_domain.rst
```diff
@@ -9,7 +9,7 @@ Clock domains
 Introduction
 ------------
 
-In *SpinalHDL*\ , clock and reset signals can be combined to create a **clock domain**. Clock domains can be applied to some areas of the design and then all synchronous elements instantiated into those areas will then **implicitly** use this clock domain.
+In SpinalHDL, clock and reset signals can be combined to create a **clock domain**. Clock domains can be applied to some areas of the design and then all synchronous elements instantiated into those areas will then `implicitly` use this clock domain.
```
wifasoi, line 12: Why use inline formatting instead of the bold text?
(I prefer the bold text, but I don't have a strong opinion about it)

```suggestion
In SpinalHDL, clock and reset signals can be combined to create a **clock domain**. Clock domains can be applied to some areas of the design and then all synchronous elements instantiated into those areas will then **implicitly** use this clock domain.
```

source/SpinalHDL/Structuring/components_hierarchy.rst
```diff
@@ -58,37 +58,34 @@ The syntax to define inputs and outputs is the following:
...
+       T should extend ``IMasterSlave`` – Some documentation is available :ref:`here <interface_eaxample_apb>`. You may not actually need the brackets, so ``master T`` is fine as well.
      - T
 
 
 There are some rules to follow with component interconnection:
 
 
-* Components can only read output and input signals of child components
-* Components can read their own output port values (unlike VHDL)
+* Components can only `read` output and input signals of child components.
```
wifasoi, line 72: As below, i personally prefer inline formatting for reference code spinets. For putting emphasis on words I would rater user the usual **bold** and _Italic_ style.

```suggestion
* Components can only **read** output and input signals of child components.
```

wifasoi (commented):
Just nitpicking stuff, as always :) (you can just dismiss my comment, or apply my suggestion. Your call).

Now it's becoming a bit of a cliché, but I still I want thank you for your contribution to the docs.

source/SpinalHDL/Structuring/components_hierarchy.rst
philipaconrad: I think in rst, single backticks \` cause *italicized* formatting, but bolds are definitely fine with me for this use case. I wanted to call some attention to that part of the sentence, because it was easy to miss when I was first skimming that page of the docs.

source/SpinalHDL/Structuring/clock_domain.rst
philipaconrad: I think I axed the bolds here because they looked inconsistent with the majority of the other pages I've edited. That said, I actually *like* the bold formatting, so I'll happily throw that back in, since I think it makes the docs easier to read.

philipaconrad:
You're welcome! :smiley: The thanks may be at risk of becoming a cliche, but it is appreciated.

I appreciate the review comments as well.

PR #54, opened 2021-02-17
Add uart example for an unreleased feature.

volatile-static:
Forgive my poor English please.


wifasoi:
Sorry for the delay.
Anyway what do you mean by unreleased?
if is so, it should  committed on the dev branch.

Need to test the example, bit they seams fine.

volatile-static:
It has been merged into dev [here](https://github.com/SpinalHDL/SpinalHDL/commit/2eb815a5bd87d0b866e3dbd85a2ca95885f8a902). And it is going to be released in v1.4.4.

volatile-static:
The generated Verilog code looks good. @wifasoi

PR #58, opened 2021-03-15
A tip about virtual pin.

volatile-static:
miaou.addAttribute("altera_attribute", "-name VIRTUAL_PIN ON")


Dolu1990:
666 is for sure a hell lot ^^

PR #59, opened 2021-03-17
add a section called "Convert Bits back to Bundle" to document the ".assignFromBits" operator

PulseRainmaker:
In the current documentation, the .assignFromBits operator is not explicitly mentioned for Bundle (Though it is briefly mentioned in the Types (Data (Bool, Bits, UInt, SInt, Enum, Bundle, Vec)). However, as the reverse of ".asBits", the ".assignFromBits" can be a useful operator.


Dolu1990:
Thanks :)

PR #71, opened 2022-01-19
binarySystem doc added

jijingg:


jijingg:
@Dolu1990 more one question,  why Auto Linkceck and buildtest always failed .  There is no problem compiling locally with command `make html `, everything is well.

wifasoi:
Ahaa, I see. Is the old friend wavedrom. I'll check it to se what's up.

Dolu1990:
@jijingg  Yes right, i don't know what happend XD
But you right, the pr is good ^^ 

Thanks !

PR #74, opened 2022-02-04
Switch Is clause factorization

ronan-lashermes:
Just showing how to add several cases in the same is-clause, separating them with a comma.


wifasoi:
Doing gods work :D
merged (still need to fix the CI/CD)

PR #75, opened 2022-02-15
Register vectors

ronan-lashermes:
Examples to create vectors of registers, with examples for simple of more complex initialization methods.


wifasoi:
As always.. need to fix the CI/CD.
LGTM

PR #76, opened 2022-02-17
clock_domains.rst: add note about register capture

makigumo:
This adds a note about register capture that was not obvious to me even after reading this chapter (https://github.com/SpinalHDL/SpinalHDL/issues/614).


Dolu1990:
Thanks :D

PR #78, opened 2022-02-25
Correction on the proper way to initialize a Flow in Register vec example

ronan-lashermes:


wifasoi:
Ok.. it's fine.. but what about using this: https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Other%20language%20features/utils.html#the-safe-way

(the most proper-er way :P)

wifasoi:
Seams fine to me. Will merge

PR #83, opened 2022-03-23
 add 'Signal priorities in HDL generation'

NomadShen:
Explain the prioirties of reset and clock enable signals in the clock domain class, and a small example is given.


Dolu1990:
Thanks ^^

Dolu1990:
A few changes : 
https://github.com/SpinalHDL/SpinalDoc-RTD/commit/75523826dc82a24a6e3f2bf3943c6192aae464c7

PR #84, opened 2022-03-25
Fix some typo and add environment variables configurations in VCS doc.

wswslzp:
Hi, @Dolu1990 

I have fixed some typos and added some environment variables configurations in VCS doc to enable the FSDB dump feature.


Dolu1990:
Thanks ^^

PR #91, opened 2022-04-20
Replace "x bit" with "x bits"

dnltz:
A lot of non data type pages still used bit instead of bits like:
UInt(8 bit). Replaced them with s/bit)/bits)/g in all files.

Signed-off-by: Daniel Schultz <dnltz@aesc-silicon.de>


Dolu1990:
Thanks ^^

PR #94, opened 2022-05-01
Add AFix documentation

kleinai:
I haven't used sphinx before so feel free to correct me.

### DIFF
--- a/source/SpinalHDL/Data types/AFix.rst
+++ b/source/SpinalHDL/Data types/AFix.rst
@@ -17,15 +17,15 @@ AFix can be created using bit sizes or exponents:
 
 .. code-block:: scala
 
-  U(12 bit)        // U12.0
-  UQ(8 bit, 4 bit) // U8.4
-  U(8 exp, 12 bit) // U8.4
+  U(12 bits)        // U12.0
+  UQ(8 bits, 4 bits) // U8.4
+  U(8 exp, 12 bits) // U8.4
   U(8 exp, -4 exp) // U8.4
   U(8 exp, 4 exp)  // U8.-4
 
-  S(12 bit)        // S11 + sign
-  SQ(8 bit, 4 bit) // S8.4 + sign
-  S(8 exp, 12 bit) // S8.3 + sign
+  S(12 bits)        // S11 + sign
+  SQ(8 bits, 4 bits) // S8.4 + sign
+  S(8 exp, 12 bits) // S8.3 + sign
   S(8 exp, -4 exp) // S8.4 + sign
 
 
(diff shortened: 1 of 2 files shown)


source/SpinalHDL/Data types/AFix.rst
```diff
@@ -0,0 +1,205 @@
...
+
+
+Declaration
+^^^^^^^^^^^
+
+AFix can be created using bit sizes or exponents:
+
+.. code-block:: scala
+
+  U(12 bits)        // U12.0
```
Dolu1990, line 20: Isn't it AFix.U ?

source/SpinalHDL/Data types/AFix.rst
```diff
@@ -0,0 +1,205 @@
...
+
+``U(8 exp, 4 exp)`` will have a range of 0 to 256
+
+
+Custom range ``AFix`` values can be created be directly instantiating the class.
+
+.. code-block:: scala
+
+  AFix(0, 4096, 0 exp)
+  AFix(-256, 256, -2 exp)
```
Dolu1990, line 48: Without comments it can be confusing, as first arg isn't the min value, but the the number of time the resolution.

source/SpinalHDL/Data types/AFix.rst
```diff
@@ -17,50 +17,52 @@ AFix can be created using bit sizes or exponents:
...
+  AFix.AFix(4096, 0, 0 exp)     // [0 to 4096, 2^0]
+  AFix.AFix(256, -256, -2 exp)  // [-256 to 256, 2^-2]
+  AFix.AFix(16, 8, 2 exp)       // [8 to 16, 2^2]
 
 
 The ``maxValue`` and ``minValue`` stores what backing integer values are representable.
 These values represent the true fixed-point value after multiplying by ``2^exp``.
 
-``U(2 exp, -1 exp)`` can represent:
+``AFix.U(2 exp, -1 exp)`` can represent:
```
Dolu1990, line 57: isn't it 0, 0.5, 1, 1.5, 2, 2.5, 3, 3.5 ?

Dolu1990:
Also in general before considering the feature as non-experimental we realy have to be sure that the API is final and smooth ^^ 

I think it is important to add a note on the top of the doc that this feature is experimental and that the api may change in future release for two reasons : 
- Prevent users from expecting a stable API, warning them.
- To allow us changing things to make things better :)

kleinai:
There's a weird issue where the new AFix section doesn't show up when looking at other sections. Maybe I'm not registering the new page somewhere?

wifasoi:
When you add a new page.. you need to do a clean build with `make clean` (because the toc is cached, and is not correctly refreshed).
Try that first, otherwise is some heading fuckery. (But I'm almost certain is a compilation cache problem)

Dolu1990:
Seems good to me ^^

@kleinai Read for merge ?

kleinai:
I would say yes but it looks like I trailed off with "Rounding is required". I'll go back and finish that paragraph later today

kleinai:
Now it's done!

source/SpinalHDL/Data types/AFix.rst
wifasoi: IS this resolved.. I don't see any comment/or modification on it. Should i start the merge? (if yes just click resolve conversation)

Dolu1990:
Thanks ^^

PR #97, opened 2022-05-05
Bunch of doc additions on basic stuff

ronan-lashermes:
- Adding a description of lib helper functions for Vec
- Adding mention of setIdle and push for Flow + example
- Adding mention of PriorityMux with the other lib muxes


wifasoi (commented):
Some rewording. Tell me if they sound better to you, and apply my suggestion at your disgression.
Some of the old one are worded as question, so they did not sound right to me.

wifasoi:
Tnaks for the PR, and making the documentation a little more complete by the day :D
Merging!

PR #98, opened 2022-05-10
example of specify a register reset-type as boot

jijingg:
sync document of discussion of https://github.com/SpinalHDL/SpinalHDL/pull/715


Dolu1990:
Thanks :)

PR #99, opened 2022-05-11
Add set as reg

numero-744:
https://github.com/SpinalHDL/SpinalHDL/issues/729


numero-744:
Sorry for the multiple pushes, I initially modified the doc from the GitHub web interface and I am not used to it, the base branch was the wrong one.

PR #100, opened 2022-05-24
Update VCS doc about new environment setup method.

wswslzp:


wifasoi:
LGTM!
thanks for the PR

PR #101, opened 2022-06-04
use pastValid but withPast which is removed.

Readon:


wifasoi:
Thanks for the pr (and sorry for the late merge).

PR #104, opened 2022-06-07
Add config example with methods and a require()

numero-744:


wifasoi:
3 in a row!
Still thanks for taking the time to improving the docs.
LGTM merging!

PR #107, opened 2022-07-27
Fix some minor problems in Libraries and Simulation

CerryXu:
Fix some minor problems in Libraries and Simulation 
val a = 32.toBigInt        a is 32
There is an error with the same code on the official website

image
change to Apb3BusInterface
Interrupt Factory There are  small problems

1. is setWhen  no setwhen
2.is  frame_int_raw no fram_int_force
3. Status No driver 
Because clk_B in StreamFifoCC does not reset, but the default clock will reset, you need to manually add withReset=false  
And the dut. PopClock. DeassertReset () and dut popClock. AssertReset () have to delete 
image

### DIFF
--- a/source/SpinalHDL/Libraries/binarySystem.rst
+++ b/source/SpinalHDL/Libraries/binarySystem.rst
@@ -217,7 +217,7 @@ BigInt enricher
 .. code-block:: scala
 
    $: 32.toBigInt
-   34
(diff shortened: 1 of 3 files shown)


wifasoi:
LGTM
Merging!
Thanks for correcting the docs! :D

PR #112, opened 2022-08-28
Small fixes for the Bits/Int pages

andreasWallner:


source/SpinalHDL/Data types/Int.rst
```diff
@@ -392,12 +392,21 @@ Misc
...
+     - Return the range (x.high downto 0)
+     - Range
+   * - x.minValue
+     - Lowest value that x can take on (e.g. 0 for UInt)
+     - BigInt
+   * - x.maxValue
+     - Highest value that x can take on
+     - BigInt
+   * - x.valueRange
+     - Return the range (x.minValue downto x.maxValue). Note can't be used for value which overflow the JVM Int capacity.
```
numero-744, line 405: - It is `to` not `downto`
- I would reformulate for those who don't know about JVM
- I am not sure if this use of "take on" is correct

What do you think about:

```
Return the range of values x can take. Bounds must not overflow a SInt(32 bits) (JVM restriction)
```

numero-744 (changes requested):

source/SpinalHDL/Data types/Int.rst
```diff
@@ -392,12 +392,21 @@ Misc
...
      - Return the least significant bit
      - Bool
-   * - x.range
-     - Return the range (x.high downto 0)
-     - Range
    * - x.high
      - Return the upper bound of the type x
      - Int
+   * - x.bitsRange
+     - Return the range (x.high downto 0)
```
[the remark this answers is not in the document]
andreasWallner: Fixed for both Bits and Int pages

source/SpinalHDL/Data types/Int.rst
andreasWallner: fixed the direction, added a clearer description that does not elude to the JVM as much, removed the "take on" wording

source/SpinalHDL/Data types/Int.rst
```diff
@@ -396,13 +396,13 @@ Misc
      - Return the range (0 to x.high)
      - Range
    * - x.minValue
-     - Lowest value that x can take on (e.g. 0 for UInt)
+     - Lowest value of x (e.g. 0 for UInt)
```
numero-744, line 399: += possible

for consistency with valueRange

source/SpinalHDL/Data types/Int.rst
```diff
@@ -396,13 +396,13 @@ Misc
...
      - BigInt
    * - x.maxValue
-     - Highest value that x can take on
+     - Highest value of x
```
numero-744, line 402: += possible

for consistency with valueRange

source/SpinalHDL/Data types/Int.rst
```diff
@@ -443,6 +443,9 @@ Misc
      - Return the absolute value of the UInt value with symmetric, shrink 1 bit
      - UInt(w(mySInt) - 1 bits)
 
+.. note::
```
numero-744, line 446: Yes, better as a rst note! :+1:

source/SpinalHDL/Data types/bits.rst
```diff
@@ -276,7 +276,7 @@ Misc
      - Return the range (0 to x.high)
      - Range
    * - x.valueRange
-     - Return the range (x.minValue downto x.maxValue). Note can't be used for value which overflow the JVM Int capacity.
+     - Return the range of valid x values, interpreted as an unsigned integer (0 to 2**width - 1).
```
numero-744, line 279: Won't `**` be interpreted by rst as bold?

The formulation can be taken from you earlier modifications:

```rst
- Return the range from minimum to maximum possible value of x (0 to 2 ** width - 1).
```

source/SpinalHDL/Data types/Int.rst
andreasWallner: done

source/SpinalHDL/Data types/Int.rst
andreasWallner: done

source/SpinalHDL/Data types/bits.rst
andreasWallner: done, I checked and `**` renders fine since it's never closed it seems, I still escaped it

wifasoi:
LGTM,
Thank you for your contribution and (@)numero-744 for doing the review.
Merging!

source/SpinalHDL/Data types/bits.rst
```diff
@@ -273,12 +273,13 @@ Misc
      - Return bitcount
      - Int
    * - x.bitsRange
-     - Return the range (x.high downto 0)
+     - Return the range (0 to x.high)
+     - Range
    * - x.valueRange
-     - Return the range (x.minValue downto x.maxValue). Note can't be used for value which overflow the JVM Int capacity.
+     - Return the range of minimum to maximum x values, interpreted as an unsigned integer (0 to 2 \*\* width - 1).
```
saahm, line 279: Three things that come to my mind here: 
1. it would be nice to keep range descriptors that are similar to the code "(x.minValue downto x.maxValue)"
2. Usually I'd expect "HI downto LO " ranges instead of "LO to HI"
3. The previous comment had this information about the overflow of the JVM Int capacity. Shouldnt that be kept as a note?

andreasWallner: ad 1: I wanted to avoid that since it's not a numeric type that we are talking about
ad 2: that not how it's implemented, a range of low to high is returned
ad 3: That comment is still there, just as a not at the end of the table, like the other caveats that are described on the page

numero-744: For the first point, the notion of "range" is [by definition](https://en.wikipedia.org/wiki/Interval_(mathematics)) about real numbers: there must be an axis / ordering relationship so in a way, "range" can exist only while considering Bits as real numbers, and in this case UInt (it could be SInt or even floating point numbers).

So IMO there is an issue in the API about the notion of range for Bits; and as far as this function exists, explicitly stating that this specific function is, in a way, casting Bits to UInt, is not an issue.

PR #117, opened 2022-09-13
Fix behavior of 'default' in when

andreasWallner:
Contrary to what was described it is in a fully covered mux not allowed to not have a default, it is prohibited.

Also do a bit of a shuffle to not understate muxListDc before having introduced muxList


Dolu1990:
Thanks :D

PR #119, opened 2022-10-05
update chinese version link and formal naming policy.

Readon:


Readon:
> Was this Chinese doc auto-translated/generated from the SpinalDoc-RTD ? Or is it maintained manualy ? Quite a big markdown ^^

It is translated from this doc, and maintained manually.

wifasoi:
I could enable multilangiage support... never tried but maybe helps with maintain 2 docs.

Dolu1990:
@wifasoi Hoo that could be great. I'm curious ^^

PR #120, opened 2022-10-09
Fix build

numero-744:
I could not build locally (Python errors), this PR fixes it.

With help of @wifasoi via Matrix channel.


Dolu1990:
Nice Thanks @numero-744 @wifasoi  :D

PR #121, opened 2022-10-13
fix: assignment overlap in Semantic/assignments

numero-744:
It would be suuuuper cool to automatically test examples of docs, [like what is done with Rustdoc](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#passing-or-failing-a-doctest) (should I open an issue with suggestions to do this?).

### DIFF
--- a/source/SpinalHDL/Semantic/assignments.rst
+++ b/source/SpinalHDL/Semantic/assignments.rst
@@ -16,21 +16,23 @@ There are multiple assignment operators:
    * - Symbol
      - Description
    * - ``:=``
-     - Standard assignment, equivalent to ``<=`` in VHDL/Verilog. The last assignment to a variable wins; the value is not updated until the next simulation delta cycle.
+     - Standard assignment, equivalent to ``<=`` in VHDL/Verilog.
    * - ``\=``
      - Equivalent to ``:=`` in VHDL and ``=`` in Verilog. The value is updated instantly in-place.
    * - ``<>``
      - Automatic connection between 2 signals or two bundles of the same type. Direction is inferred by using signal direction (in/out). (Similar behavior to ``:=``\ )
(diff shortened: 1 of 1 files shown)


Dolu1990:
Ahh why removing "The last assignment to a variable wins; the value is not updated until the next simulation delta cycle." ?

About running example, yes that would be cool ^^
I'm just scared that Scala + JVM is sooo  laggy to run, it would take ages XD

numero-744:
Sorry, forgot to mention where this modification comes from: the really end of this comment: https://github.com/SpinalHDL/SpinalHDL/issues/895#issuecomment-1277437886

Let's split the sentence at `;`:

First part, because it is not true without a `when` (explanation has been added below the table, in the text: it triggers assignment errors).

The text after the `;` could be kept, but the comparison with VHDL / Verilog is IMO enough for people who take care of this ^^'.

About running examples, it is just like unit-testing, but with pre-processing step (let's discuss it in an issue I'm gonna open :smile:).

Dolu1990:
>  but the comparison with VHDL / Verilog is IMO enough for people who take care of this ^^'.

One day, one this universe or another one, people may learn SpinalHDL as their first hardware description tool :D That's why keeping things to not refer too much to those language isn't a bad thing.
(that's the case for quite a few software dev people interrested into hardware)

numero-744:
My point is that "Simulation delta cycle" would probably not be a thing if VHDL and Verilog were not used for hardware description, because it is a simulation notion for a simulation language, that VHDL and Verilog are, and SpinalHDL is not.

I mean it could probably exist in simulator internals, but designers should not be aware of this anyway because they should think "wires", "registers" and "logic", not "simulation processes" and "delta cycles", don't you think?

PR #122, opened 2022-10-13
remove unused files to remove warnings

numero-744:
Unused files are in git history so they can be restored later if needed.


Dolu1990:
De l'achéologie ^^

PR #134, opened 2022-10-28
fix: last example "Simulation support"

numero-744:
It was not building. Now it builds but it still does not pass the test.

The test fails: `Timeout trigger after 10000 units of time`


Dolu1990:
I will check why it isn't passing

Dolu1990:
For me it is fine : 
[Done] Simulation done in 31.952 ms

numero-744:
For me too it worked the first time, maybe an issue about randomization?

numero-744:
Watch this seed: 172735708

numero-744:
I have added a `println` in each `StreamMonitor`, it seems to push but never pop.

numero-744:
Merge if okay for you

Dolu1990:
Ahhh nice catch, yes flush randomization issue ^^
Thanks

Issue #138, opened 2022-11-07
Document switch(coverUnreachable = ...)


numero-744:
Source file to modify: [source/SpinalHDL/Semantic/when_switch.rst](https://github.com/SpinalHDL/SpinalDoc-RTD/blob/master/source/SpinalHDL/Semantic/when_switch.rst)

See https://github.com/SpinalHDL/SpinalHDL/issues/946

mrcmry:
Closed as completed.

PR #139, opened 2022-11-09
document coverUnreachable and strict options for the switch block

saahm:
For #138 discussed in [SpinalHDL/SpinalHDL#946](https://github.com/SpinalHDL/SpinalHDL/issues/946) 

also added documentation for the strict option since its part of the same structure and would belong in the same place.

I am open for slightly less technical/deep descriptions of these parameters as I dont know if it would confuse anyone reading that first time.

### DIFF
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -101,7 +101,7 @@ is equivalent to
 Additional options
 ^^^^^^^^^^^^^^^^^^
 
-Sometimes handling all cases can become unwieldy and error prone so SpinalHDL, by default, will handle the uncovered cases with the last ``is`` block.
+Sometimes handling all cases can become unwieldy and error prone so SpinalHDL, by default, handles the uncovered cases with the last ``is`` block.
 To explicitly declare and define a ``default`` block the option ``coverUnreachable`` can be passed to the switch
 
 .. code-block:: scala
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -141,7 +141,7 @@ To relax the strictness of the ``switch`` elaboration the ``strict = false`` can
           foo := 4
           bar := 2
       }
-      is(OP_SLT){
+      is(OP_SLT) {
           foo := 2
           bar := 8
       }
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -101,7 +101,7 @@ is equivalent to
 Additional options
 ^^^^^^^^^^^^^^^^^^
 
-By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if a switch/is contains a ``default`` statement while already all the possible logical values of the 'switch' are covered by the 'is' definitions. You can drop this error reporting by specifying `` switch(myValue, coverUnreachable = true){ ... }``.
+By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if a ``switch`` contains a ``default`` statement while all the possible logical values of the ``switch`` are already covered by the ``is`` statements. You can drop this error reporting by specifying `` switch(myValue, coverUnreachable = true) { ... }``.
 
 .. code-block:: scala
   
@@ -110,20 +110,22 @@ By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if 
       is(1) { ... } 
       is(2) { ... }
       is(3) { ... }
-      default { ... } //This will be ok
+      default { ... } // This will be okay
   }
   
-Note, this check is done on the logical values, not on the physical values. For instance, if you have a SpinalEnum(A,B,C) encoded in a on-hot manner, SpinalHDL will only care about the A,B,C values ("001" "010" "100"). Pyhsical values as "000" "011" "101" "110" "111" will not be taken in account.
+.. note::
+
+   This check is done on the logical values, not on the physical values. For instance, if you have a SpinalEnum(A,B,C) encoded in a on-hot manner, SpinalHDL will only care about the A,B,C values ("001" "010" "100"). Pyhsical values as "000" "011" "101" "110" "111" will not be taken in account.
 
 
-By default, SpinalHDL will generate an "DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT" error if a given ``is`` statements provides multiple time the same value. For instance ``is(42,42) { .. }`` 
-You can drop this error reporting by specifying `` switch(myValue, strict = true){ ... }``. SpinalHDL will then take care of removing duplicated values.
+By default, SpinalHDL will generate a "DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT" error if a given ``is`` statement provides multiple times the same value. For instance ``is(42,42) { ... }`` 
+You can drop this error reporting by specifying ``switch(myValue, strict = true){ ... }``. SpinalHDL will then take care of removing duplicated values.
 
 .. code-block:: scala
   
   switch(value, strict = false) {
-      is(0)  { ... }
-      is(1,1,1,1,1)  { ... } //This will be ok
+      is(0) { ... }
+      is(1,1,1,1,1) { ... } // This will be okay
       is(2) { ... }
   }
 
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -112,6 +112,8 @@ By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if 
       is(3) { ... }
       default { ... } //This will be ok
   }
+  
+Note, this check is done on the logical values, not on the physical values. For instance, if you have a SpinalEnum(A,B,C) encoded in a on-hot manner, SpinalHDL will only care about the A,B,C values ("001" "010" "100"). Pyhsical values as "000" "011" "101" "110" "111" will not be taken in account.
 
 
 By default, SpinalHDL will generate an "DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT" error if a given ``is`` statements provides multiple time the same value. For instance ``is(42,42) { .. }`` 
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -137,7 +137,7 @@ To relax the strictness of the ``switch`` elaboration the ``strict = false`` can
   val bar = UInt(8 bits)
   switch(io.instruction, strict = false) {
       // 
-      is(OP_ADD, OP_SUB){
+      is(OP_ADD, OP_SUB) {
           foo := 4
           bar := 2
       }
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -101,8 +101,7 @@ is equivalent to
 Additional options
 ^^^^^^^^^^^^^^^^^^
 
-In some cases when covering all cases of the input encoding still leaves uncovered codes. 
-In the generated HDL these will be handled by default with the last ``is`` block.
+Sometimes handling all cases can become unwieldy and error prone so SpinalHDL, by default, will handle the uncovered cases with the last ``is`` block.
 To explicitly declare and define a ``default`` block the option ``coverUnreachable`` can be passed to the switch
 
 .. code-block:: scala
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -149,7 +149,7 @@ To relax the strictness of the ``switch`` elaboration the ``strict = false`` can
           foo := 2
           bar := 8
       }
-      default{
+      default {
           foo := 0
           bar := 0
       }
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -101,58 +101,28 @@ is equivalent to
 Additional options
 ^^^^^^^^^^^^^^^^^^
 
-Sometimes handling all cases can become unwieldy and error prone so SpinalHDL, by default, handles the uncovered cases with the last ``is`` block.
-To explicitly declare and define a ``default`` block the option ``coverUnreachable`` can be passed to the switch
-
-.. code-block:: scala
-
-  switch(aluop, coverUnreachable = true) {
-    is(ALUOp.add, ALUOp.slt, ALUOp.sltu) {
-        immediate := instruction.immI.signExtend
-    }
-    is(ALUOp.sll, ALUOp.sra) {
-        immediate := instruction.shamt
-    }
-    default{
-        immediate := 0
-    }
-  }
-
-If the used values of ``ALUOp`` are all available elements of the SpinalEnum type, then without the ``coverUnreachable = true`` option SpinalHDL would throw an UNREACHABLE STATEMENT error upon elaboration.
-Without the ``default`` block the encoding for ``ALUOp.sll`` and ``ALUOp.sra`` are the default cases in the generated HDL.
-
-When using defined constants to compare against within a ``switch`` block it can occasionally happen that duplicates within a ``is`` value occur. 
-By default duplicates of ``is`` conditions are not ignored but identified as an error. 
-To relax the strictness of the ``switch`` elaboration the ``strict = false`` can be passed (by default ``strict = true`` thus preventing duplicate ``is`` conditions).
+By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if a switch/is contains a ``default`` statement while already all the possible logical values of the 'switch' are covered by the 'is' definitions. You can drop this error reporting by specifying `` switch(myValue, coverUnreachable = true){ ... }``.
 
 .. code-block:: scala
   
-  // OP_ADD and OP_SUB share the same code
-  def OP_ADD = M"000"
-  def OP_SUB = M"000"
-  def OP_SLT = M"001"
-  def OP_JMP = M"010"
-  def OP_BRK = M"101"
-  val foo = UInt(8 bits)
-  val bar = UInt(8 bits)
-  switch(io.instruction, strict = false) {
-      // 
-      is(OP_ADD, OP_SUB) {
-          foo := 4
-          bar := 2
-      }
-      is(OP_SLT) {
-          foo := 2
-          bar := 8
-      }
-      is(OP_JMP, OP_BRK) {
-          foo := 2
-          bar := 8
-      }
-      default {
-          foo := 0
-          bar := 0
-      }
+  switch(my2Bits, coverUnreachable = false) {
+      is(0) { ... }
+      is(1) { ... } 
+      is(2) { ... }
+      is(3) { ... }
+      default { ... } //This will be ok
+  }
+
+
+By default, SpinalHDL will generate an "DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT" error if a given ``is`` statements provides multiple time the same value. For instance ``is(42,42) { .. }`` 
+You can drop this error reporting by specifying `` switch(myValue, strict = true){ ... }``. SpinalHDL will then take care of removing duplicated values.
+
+.. code-block:: scala
+  
+  switch(value, strict = false) {
+      is(0)  { ... }
+      is(1,1,1,1,1)  { ... } //This will be ok
+      is(2) { ... }
   }
 
 
--- a/source/SpinalHDL/Semantic/when_switch.rst
+++ b/source/SpinalHDL/Semantic/when_switch.rst
@@ -97,6 +97,66 @@ is equivalent to
     }
   }
 
+
+Additional options
+^^^^^^^^^^^^^^^^^^
+
+In some cases when covering all cases of the input encoding still leaves uncovered codes. 
+In the generated HDL these will be handled by default with the last ``is`` block.
+To explicitly declare and define a ``default`` block the option ``coverUnreachable`` can be passed to the switch
+
+.. code-block:: scala
+
+  switch(aluop, coverUnreachable = true) {
+    is(ALUOp.add, ALUOp.slt, ALUOp.sltu) {
+        immediate := instruction.immI.signExtend
+    }
+    is(ALUOp.sll, ALUOp.sra) {
+        immediate := instruction.shamt
+    }
+    default{
+        immediate := 0
+    }
+  }
+
+If the used values of ``ALUOp`` are all available elements of the SpinalEnum type, then without the ``coverUnreachable = true`` option SpinalHDL would throw an UNREACHABLE STATEMENT error upon elaboration.
+Without the ``default`` block the encoding for ``ALUOp.sll`` and ``ALUOp.sra`` are the default cases in the generated HDL.
+
+When using defined constants to compare against within a ``switch`` block it can occasionally happen that duplicates within a ``is`` value occur. 
+By default duplicates of ``is`` conditions are not ignored but identified as an error. 
+To relax the strictness of the ``switch`` elaboration the ``strict = false`` can be passed (by default ``strict = true`` thus preventing duplicate ``is`` conditions).
+
+.. code-block:: scala
+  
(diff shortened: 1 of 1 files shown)


source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,66 @@ is equivalent to
     }
   }
 
+
+Additional options
+^^^^^^^^^^^^^^^^^^
+
+In some cases when covering all cases of the input encoding still leaves uncovered codes. 
+In the generated HDL these will be handled by default with the last ``is`` block.
```
wifasoi, line 105: The first sentence is a bit difficult to "parse", what about:
```suggestion
Sometimes handling all cases can become unwieldy and error prone so SpinalHDL, by default, will handle the uncovered cases with the last ``is`` block.
```

wifasoi (commented):
Just suggesting some rewording here. Please choose what you prefer between my suggestion, and yours. otherwise LGTM

saahm:
> Just suggesting some rewording here. Please choose what you prefer between my suggestion, and yours. otherwise LGTM

I like that very much, I was aware it was a bit difficult. But I thought any attempt is better than none and someone can help me with that :) Glad for the suggestion

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
     }
   }
 
+
+Additional options
+^^^^^^^^^^^^^^^^^^
+
+Sometimes handling all cases can become unwieldy and error prone so SpinalHDL, by default, will handle the uncovered cases with the last ``is`` block.
```
numero-744, line 104:
```suggestion
Sometimes handling all cases can become unwieldy and error prone so SpinalHDL, by default, handles the uncovered cases with the last ``is`` block.
```

It is implemented and not to be implemented, so I think using present of general truths here is clearer.

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
...
+To explicitly declare and define a ``default`` block the option ``coverUnreachable`` can be passed to the switch
```
numero-744, line 105: I am not sure to understand correctly what it is about here so I suggest an improvement.

If I understand correctly there are two kinds of "cover" to do (I use my words here but feel free to use other words if they are clearer):

* "logic" cover: the possible values that can be applied on the wires from the logic (enumerated types)
* "bits" cover: all possible combination of 0s and 1s in the wires. It is a superset of "logic" cover.

For example, using this encoding in an enumerated type (maybe provide an example with actual code)

MOV -> b00
ADD -> b10
SUB -> b11

"logic" cover requires covering the 3 cases MOV, ADD and SUB and "bits" cover requires covering the 4 cases b00 to b11, which are the 3 cases from "logic" cover + unreachable case b01.

By default SpinalHDL only checks "logic" coverage. To reach full "bits" coverage, unreachable cases are redirected to the last case of the `switch` (note: it can be an `is` or a `default`).

If you have defined `is` cases with a complete "logic" cover and want to define a specific behavior for unreachable cases, you will add a `default` case; but as all reachable cases were already covered, it will be considered unreachable. To not consider it unreachable and use it for unreachable cases, use `coverUnreachable = true`.

Example…

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
...
+  }
+
+If the used values of ``ALUOp`` are all available elements of the SpinalEnum type, then without the ``coverUnreachable = true`` option SpinalHDL would throw an UNREACHABLE STATEMENT error upon elaboration.
+Without the ``default`` block the encoding for ``ALUOp.sll`` and ``ALUOp.sra`` are the default cases in the generated HDL.
+
+When using defined constants to compare against within a ``switch`` block it can occasionally happen that duplicates within a ``is`` value occur. 
+By default duplicates of ``is`` conditions are not ignored but identified as an error. 
+To relax the strictness of the ``switch`` elaboration the ``strict = false`` can be passed (by default ``strict = true`` thus preventing duplicate ``is`` conditions).
+
+.. code-block:: scala
```
numero-744, line 128: I do not understand the example, can you explain the consequence of using `strict = false` here, please? (Behavior with vs without it)

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
...
+  
+  // OP_ADD and OP_SUB share the same code
+  def OP_ADD = M"000"
+  def OP_SUB = M"000"
+  def OP_SLT = M"001"
+  def OP_JMP = M"010"
+  def OP_BRK = M"101"
+  val foo = UInt(8 bits)
+  val bar = UInt(8 bits)
+  switch(io.instruction, strict=false){
```
numero-744, line 138:
```suggestion
  switch(io.instruction, strict = false) {
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
...
+      // 
+      is(OP_ADD, OP_SUB){
```
numero-744, line 140:
```suggestion
      is(OP_ADD, OP_SUB) {
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
...
+          foo := 4
+          bar := 2
+      }
+      is(OP_SLT){
```
numero-744, line 144:
```suggestion
      is(OP_SLT) {
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
...
+          foo := 2
+          bar := 8
+      }
+      is(OP_JMP, OP_BRK){
```
numero-744, line 148:
```suggestion
      is(OP_JMP, OP_BRK) {
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,65 @@ is equivalent to
...
+          foo := 2
+          bar := 8
+      }
+      default{
```
numero-744, line 152:
```suggestion
      default {
```

numero-744 (commented):
I have never used SpinalHDL enumerated types yet (except indirectly via FSM) so I am discovering things here and I review this PR as a newbie

source/SpinalHDL/Semantic/when_switch.rst
saahm: So if I leave out `strict = false` what happens is that SpinalHDL gives an error (`DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT`). Using this option relaxes the elaboration of the switch block to allow the duplicate element (and only keep one in the generated Verilog (as otherwise Verilator would complain too -- I remember talking about this to Dolu when I got this issue the first time, I dont recall if I created an issue or just asked in the gitter).

source/SpinalHDL/Semantic/when_switch.rst
saahm: Ah, so it matters if I use an enumerate (logic cover) or just bits (bits cover) [regardless of constants or immediate values].
I will try to rewrite it then to make it a bit more close to that understanding. Maybe I can also come up with a good example to code show this.

numero-744: Well, it is just my understanding and I do not use it so maybe my understanding is wrong.

@Dolu1990

Dolu1990: >  by default, will handle the uncovered cases with the last ``is`` block.

That sound wrong or else confusing to me. 

The goal here is to document coverUnreachable right ?

What's about something kind of like :

"""
If a switch/is contains a 'default' statement while already all the possible values of the 'switch' are covered by the 'is' values, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error. You can drop this error reporting by specifying switch(myValue, coverUnreachable = true){ ... }. 
"""

?

numero-744: @Dolu1990 was my suggestion correct?

numero-744: OH is a good example too!

Hmm so my question is (with your example which is more natural :smile:): what if I have several 1's in a OH? How is it handled by the switch? My understanding of what was suggested in this PR is that the last block (`is` or `default`) is used for that. Is this it?

source/SpinalHDL/Semantic/when_switch.rst
numero-744: So which one is used? The last one (last-assignment-wins rule)?

I suggest this kind of structure, which is more concise and explains the example.

```rst
By default, duplicates of ``is`` conditions are identified as an error.
This strictness can be released passing ``strict = false`` to ``switch``, then XXX happens.
In the example below, Xxx and Xxx are duplicates, and ``strict = false`` makes XXX happen when XXX.
```

source/SpinalHDL/Semantic/when_switch.rst
Dolu1990: Hmm how do you define a OH in a switch statement ?

saahm: I guess some IO thats supposed to be OH encoded but has a 1s-count > 1, (1000100) instead of (1000000). 
I would think that switch statements (in Verilog) usually go to 'default' that.

Dolu1990: Issue with putting a is(000001000) is that is consume more logic than required, should idealy only check is(M"-------1-----"), and be exclusive against  other is(xxx) in the same switch. But not sure how to properly generate that verilog

numero-744:
```scala
 class Dut extends Component {
    val io = new Bundle {
      val i = in Bits (5 bits)
      val o = out Bits (3 bits)
    }
    switch(io.i, coverUnreachable = true) {
      is(M"----1") { io.o := 0 }
      is(M"---1-") { io.o := 1 }
      is(M"--1--") { io.o := 2 }
      is(M"-1---") { io.o := 3 }
      is(M"1----") { io.o := 4 }
      //default { io.o := 7 }
    }
  }
```

With `default` commented, I get LATCH DETECTED

With `default` uncommented, I get verilator error about two many errors.

The with vs without `coverUnreachable` does not change anything in both cases.

If I add `is(0) { io.o := 7 }` I also get LATCH DETECTED (sounds like a bug?).

We are talking about using literals but didn't you confirm me earlier this option was only about enumerated types?

saahm: Thats an interesting example there why the generation has issues. 

Technically it should recognize for the OH, but thats not "obvious" once someone has some don't cares in their bitmasks to check against.
Usually it should be at least possible if a default case is provided (talking about how the Verilog could look like).

@Dolu1990 i think then it has to be clarified too if and how to use don't care and OH for switches? I'm not sure, it seems like some niche cases here but they should be considered too in the documentation (this should maybe be a different PR tho)

Dolu1990: @numero-744 @saahm i think the "intended" way is to generate zero logic overhead while staying very explicit and not inferring a latch is : 

```scala
    switch(io.i) {
      is(M"----1") { io.o := 0 }
      is(M"---1-") { io.o := 1 }
      is(M"--1--") { io.o := 2 }
      is(M"-1---") { io.o := 3 }
      is(M"1----") { io.o := 4 }
      default{ io.o.assignDontCare()}
    }
```

or 

```scala
    io.o.assignDontCare()
    switch(io.i) {
      is(M"----1") { io.o := 0 }
      is(M"---1-") { io.o := 1 }
      is(M"--1--") { io.o := 2 }
      is(M"-1---") { io.o := 3 }
      is(M"1----") { io.o := 4 }
    }
```

> If I add is(0) { io.o := 7 } I also get LATCH DETECTED (sounds like a bug?).

Not realy a bug, more like a limitation of how SpinalHDL is checking coverage (which is kinda a basic implementaion XD)

Dolu1990: By the way, i forgot that SpinalHDL supported is(M"----1") natively in verilog (5'b????1 : begin) XD

numero-744: ### Using masking

As there is the 0 case, I was wondering why it didn't work with the `default`, and the issue seems to be the generated code:

```systemverilog
  always @(*) begin
    casez(io_i)
      5'b????1 : begin
        io_o = 3'b000;
      end
      5'b???1? : begin
        io_o = 3'b001;
      end
      5'b??1?? : begin
        io_o = 3'b010;
      end
      5'b?1??? : begin
        io_o = 3'b011;
      end
      5'b1???? : begin
        io_o = 3'b100;
      end
      default : begin
        io_o = 3'b111;
      end
    endcase
  end
```

```
Case values overlap (example pattern 0x3)
   18 |       5'b???1? : begin
      |       ^~~~~~~~
```

For instance it works with:

```scala
      is(M"----1") { io.o := 0 }
      is(M"---10") { io.o := 1 }
      is(M"--100") { io.o := 2 }
      is(M"-1000") { io.o := 3 }
      is(M"10000") { io.o := 4 }
      default { io.o := 7 }
```

But it does not answer the question (@Dolu1990 is it what you meant by "But not sure how to properly generate that verilog"?)

Note: it seems correct according to https://www.verilogpro.com/verilog-case-casez-casex/ maybe this lint should be disabled? Also, wouldn't just using `if` matching be okay?

Anyway it seems off-topic as `coverUnreachable` is about enumerated types. The actual question here is, using an enumerated type with OH encoding, what if the input has two 1s.

### Using enumerated type

So I have generated verilog:

```scala
  object MyEnum extends SpinalEnum(binaryOneHot) {
    val zero, one, two, three, four = newElement()
  }
  class Dut extends Component {
    val io = new Bundle {
      val i = in port MyEnum
      val o = out port Bits(3 bits)
    }
    switch(io.i) {
      import MyEnum._
      is(zero) { io.o := 0 }
      is(one) { io.o := 1 }
      is(two) { io.o := 2 }
      is(three) { io.o := 3 }
      is(four) { io.o := 4 }
    }
  }
```

```systemveriog
  always @(*) begin
    (* parallel_case *)
    case(1) // synthesis parallel_case
      (((io_i) & MyEnum_zero) == MyEnum_zero) : begin
        io_o = 3'b000;
      end
      (((io_i) & MyEnum_one) == MyEnum_one) : begin
        io_o = 3'b001;
      end
      (((io_i) & MyEnum_two) == MyEnum_two) : begin
        io_o = 3'b010;
      end
      (((io_i) & MyEnum_three) == MyEnum_three) : begin
        io_o = 3'b011;
      end
      default : begin
        io_o = 3'b100;
      end
    endcase
  end
```

Then I replace one case by `default` and I get the conclusion:

* If there is a `default` in the `switch` then unreachable cases fall into this `default`
* else unreachable cases fall into the last `is`

I think this PR can be updated with this data.

numero-744: FYI about masking: https://github.com/SpinalHDL/SpinalHDL/issues/468

numero-744: ### OH masking by-hand

Ahh ok it is to have parallel muxing without priority.

I have found more reading about it:

* [Verilog full case and parallel case](http://referencedesigner.com/tutorials/verilog/verilog_21.php)
* [One hot encoding for state machines](http://asics.chuckbenz.com/detailed_one_hot.htm)

The second one suggests:

```systemverilog
case (1'b1) // synthesis full_case parallel_case
  state[0]: begin ... end
  state[1]: begin ... end
  state[2]: begin ... end
  state[3]: begin ... end
endcase
```

I think it is like what the enumerated type generates but without the `== value`

```systemverilog
  always @(*) begin
    (* parallel_case *)
    case(1) // synthesis parallel_case
      (((io_i) & MyEnum_zero) == MyEnum_zero) : begin
        io_o = 3'b000;
      end
      ...
      // becomes
      ((io_i) & MyEnum_zero) : begin
        io_o = 3'b000;
      end
      ...
      // No default
      /*default : begin
        io_o = 3'b100;
      end*/
    endcase
  end
```

### Enumerated type

> > The actual question here is, using an enumerated type with OH encoding, what if the input has two 1s.
>
> I guess that then it is mostly undefined behaviour in verilog, and what should be done to protect against it is :

I give an answer in the second part of my comment, it was just to introduce it / a transition; see the end of the same comment.

> Which rule will enforce that ?
> Maybe the best is to specify "undefined behaviour" and that the best is for the the user to implement a fallback as "when(!isOh(myOh)) { } ?

You answer to a section about Spinal enumerated types, with manual OH encoding so you fall off-topic here XD
What happens is that SpinalHDL generates a `default` case for the last `is(…) { … }` if there is no `default { … }` (instead of a case for `MyEnum_four` in the example) and I guess synthesis tools redirect all other cases to the default case. XD

### About using `.mux`

> What's about B(4) ?

Better, thanks!

> Which kind of option exactly ?

`coverUnreachable`. We could also mention `strict` here I guess.

Dolu1990: > What happens is that SpinalHDL generates a default case for the last is(…) { … } if there is no default { … }

Ahhhhh yes right XD

> coverUnreachable. We could also mention strict here I guess.

Currently, there is no such thing for mux.

Dolu1990: > What happens is that SpinalHDL generates a default case for the last is(…) { … } if there is no default { … }

I would say, this should not be documented, as it is mostly a generation artefact, not realy a feature.

Dolu1990: Also, notes SpinalHDL replace the last "is" with a "default" only if all the logical values are covered by the "is" statements

numero-744: > I would say, this should not be documented, as it is mostly a generation artefact, not realy a feature.

I 80% agree so I agree XD

>Also, notes SpinalHDL replace the last "is" with a "default" only if all the logical values are covered by the "is" statements

Okay now I 100% agree to just don't tell about it then XD XD

So for the explanation of `coverUnreachable`, I suggest:

1. Explain logic vs bits coverage
2. Mention that to manage bits coverage a user-defined way (do not mention what Spinal does by default) if all logic coverage is done with only `is` statements, have to use `coverUnreachable` with a `default` statement.

Are you okay with it @Dolu1990 and @saahm ?

numero-744: No worries :smile:

Dolu1990: @numero-744 Seems good to me ^^

saahm: So i had time now to read a bit here, what I conclude is the documentation wont talk about all of the details we discussed about here *but* what @numero-744 concluded "show and explain logic(enum?) vs bits" and the small thing about coverage of all logic cases by using coverUnreachable for some corner cases of switch-is

Dolu1990:
I pushed some change. I'm sorry, i was wrong about the strict option. It is a option to remove duplication inside a is statement, not between is statements XD

let's me know if all is good for you with the current state of the PR.

saahm:
Works for me

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,37 @@ is equivalent to
     }
   }
 
+
+Additional options
+^^^^^^^^^^^^^^^^^^
+
+By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if a switch/is contains a ``default`` statement while already all the possible logical values of the 'switch' are covered by the 'is' definitions. You can drop this error reporting by specifying `` switch(myValue, coverUnreachable = true){ ... }``.
```
numero-744, line 104:
```suggestion
By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if a ``switch`` contains a ``default`` statement while all the possible logical values of the ``switch`` are already covered by the ``is`` statements. You can drop this error reporting by specifying `` switch(myValue, coverUnreachable = true) { ... }``.
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,37 @@ is equivalent to
...
+  
+  switch(my2Bits, coverUnreachable = false) {
+      is(0) { ... }
+      is(1) { ... } 
+      is(2) { ... }
+      is(3) { ... }
+      default { ... } //This will be ok
+  }
+  
+Note, this check is done on the logical values, not on the physical values. For instance, if you have a SpinalEnum(A,B,C) encoded in a on-hot manner, SpinalHDL will only care about the A,B,C values ("001" "010" "100"). Pyhsical values as "000" "011" "101" "110" "111" will not be taken in account.
```
numero-744, line 116:
```suggestion
.. note::

   This check is done on the logical values, not on the physical values. For instance, if you have a SpinalEnum(A,B,C) encoded in a on-hot manner, SpinalHDL will only care about the A,B,C values ("001" "010" "100"). Pyhsical values as "000" "011" "101" "110" "111" will not be taken in account.
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,37 @@ is equivalent to
...
+
+
+By default, SpinalHDL will generate an "DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT" error if a given ``is`` statements provides multiple time the same value. For instance ``is(42,42) { .. }``
```
numero-744, line 119:
```suggestion
By default, SpinalHDL will generate a "DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT" error if a given ``is`` statement provides multiple times the same value. For instance ``is(42,42) { ... }`` 
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,37 @@ is equivalent to
...
+      is(2) { ... }
+      is(3) { ... }
+      default { ... } //This will be ok
+  }
+  
+Note, this check is done on the logical values, not on the physical values. For instance, if you have a SpinalEnum(A,B,C) encoded in a on-hot manner, SpinalHDL will only care about the A,B,C values ("001" "010" "100"). Pyhsical values as "000" "011" "101" "110" "111" will not be taken in account.
+
+
+By default, SpinalHDL will generate an "DUPLICATED ELEMENTS IN SWITCH IS(...) STATEMENT" error if a given ``is`` statements provides multiple time the same value. For instance ``is(42,42) { .. }`` 
+You can drop this error reporting by specifying `` switch(myValue, strict = true){ ... }``. SpinalHDL will then take care of removing duplicated values.
```
numero-744, line 120:
```suggestion
You can drop this error reporting by specifying ``switch(myValue, strict = true){ ... }``. SpinalHDL will then take care of removing duplicated values.
```

Ah so it is `strict = true` to remove the error, not `strict = false`

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,37 @@ is equivalent to
...
+
+.. code-block:: scala
+  
+  switch(value, strict = false) {
+      is(0)  { ... }
+      is(1,1,1,1,1)  { ... } //This will be ok
+      is(2) { ... }
+  }
```
numero-744, line 128:
```suggestion
  switch(value, strict = false) {
      is(0) { ... }
      is(1,1,1,1,1) { ... } // This will be okay
      is(2) { ... }
  }
```

source/SpinalHDL/Semantic/when_switch.rst
```diff
@@ -97,6 +97,37 @@ is equivalent to
...
+By default, SpinalHDL will generate an "UNREACHABLE DEFAULT STATEMENT" error if a switch/is contains a ``default`` statement while already all the possible logical values of the 'switch' are covered by the 'is' definitions. You can drop this error reporting by specifying `` switch(myValue, coverUnreachable = true){ ... }``.
+
+.. code-block:: scala
+  
+  switch(my2Bits, coverUnreachable = false) {
+      is(0) { ... }
+      is(1) { ... } 
+      is(2) { ... }
+      is(3) { ... }
+      default { ... } //This will be ok
```
numero-744, line 113:
```suggestion
      default { ... } // This will be okay
```

Dolu1990:
Thanks @saahm @numero-744  :D

PR #143, opened 2022-11-16
Update readme

numero-744:
IMO for a readme, Markdown would be easier to use XD

EDIT: but it is easier to copy-paste from/to the source files :smile:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,16 @@
 SpinalDoc
 =========
 
+This is the documentation repository for
+`SpinalHDL <https://github.com/SpinalHDL/SpinalHDL>`_.
+
+It is published on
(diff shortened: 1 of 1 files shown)


Dolu1990:
i think we should keep some doc about how to generate the doc without docker ^^

numero-744:
@Dolu1990 I think it is still in [the README](https://github.com/SpinalHDL/SpinalDoc-RTD/blob/572c68518ce5221b37bdcf32b1d9b7266787e37c/README.rst). There is venv, Docker and native.

numero-744:
The readme? It is an RST file. A Md file would work to as it is not used by Sphinx?

Readon:
It's not abort by the compilation of rst.
it's about finding the link, as [action](https://github.com/SpinalHDL/SpinalDoc-RTD/actions/runs/3488477940/jobs/5837453219) shows.?

Dolu1990:
Hoo right, i mixed deployement with offline generation XD
Thanks

Issue #131, opened 2022-10-24
Add section for idioms


numero-744:
Spinal enables representing the code in a way which is not possible in RTL.

This section would show idioms which are useful in Spinal. For each idiom, it would explain it, how to use it, when and why it is good to use it, and maybe how it works.

Example of idiom:

```scala
val doSomethingFlag = False
def doSomething(): Unit = doSomethingFlag := True
```

I find it useful in `fsm.State.whenActive` to express the output logic.

```scala
  val fsm = new StateMachine {
    val IDLE, Rnd0, RndI, RndN, DONE = new State
    setEntry(IDLE)

    // Transition logic
    IDLE.whenIsActive { when(io.start) { goto(Rnd0) } }
    Rnd0.whenIsActive { goto(RndI) }
    RndI.whenIsActive { when(round.isLast) { goto(RndN) } }
    RndN.whenIsActive { goto(DONE) }
    DONE.whenIsActive { when(!io.start) { goto(IDLE) } }

    // Output logic, using only this idiom
    IDLE.whenIsActive { round.resetCount }
    Rnd0.whenIsActive { run; useInputs }
    RndI.whenIsActive { run }
    RndN.whenIsActive { run; withoutMixColumns }
    DONE.whenIsActive { enableOutput }
  }
```

Dolu1990:
Yes, such section would be usefull, showing all kind of funny / creative thing

Readon:
Nice Work!
This will be a shining pool of sparks for SpinalHDL.

Dolu1990:
Maybe we can do a call to discuss about patterns ?

numero-744:
I was thinking that each pattern would be introduced by a PR into the which we could discuss but a call might be good too.

Dolu1990:
sure.

numero-744:
Note for when I will have created this section:

Function-like components: a companion object with an `apply`.

```scala
  object MyComponent {
    def apply(i: SpinalEnumCraft[MyEnum.type]): Bits = {
      val myComponent = new MyComponent
      myComponent.io.i := i
      myComponent.io.o
    }
  }
```

numero-744:
Once idioms are defined we can provide snippets (for instance providing them in the template) for IDEs. For instance:

```json
	"Doing flag": {
		"scope": "scala",
		"prefix": "doing",
		"body": ["val ${2:${1}ing} = False", "def $1(): Unit = $2 := True", ""]
	}
```

[...]

PR #148, opened 2022-11-16
Convert examples

andreasWallner:
As discussed today I move this PR here.

It moves all of the code from the example sections to a project in the repo.

What was done so far:
- Removed .travis file since that service has been down for a while
- Moved code
- Fixed all small issues that blocked compilation or generation
- No major changes in the examples (only some `class` -> `case class` conversion + Fix of the odd typo

What remains WIP:
- Switch compilation to "nightly" published SpinalHDL (once that is available)
- Integrate into build (some things to be discussed, e.g. how to run generate all examples)
- Format everything consistently, I'd tend to use the SpinalHDL `scalafmt` config for that but would check first, some formatting could make it harder to read once integrated into the documentation.

Some small things that I found:
- Its often advisable to include e.g. a `(` in the start/end strings to prevent getting the wrong code block if e.g. case classes start the same (e.g. UartConfig and Uart)
- The marker search ignores leading spaces, so sadly `}` can't be used to get e.g. a full class if there is any such bracket before.

Thanks to wifasoi for triggering this, but I removed his initial scripted code extractions since: they are still available in his repo and end up being removed when merging stuff together.


numero-744:
For reference:

* nightly: https://github.com/SpinalHDL/SpinalHDL/issues/800 and https://github.com/SpinalHDL/SpinalHDL/discussions/793

examples/src/main/scala/spinaldoc/examples/simple/Apb3.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  override def asMaster(): Unit = {
+    out(PADDR,PSEL,PENABLE,PWRITE,PWDATA)
+    in(PREADY,PRDATA)
+    if(config.useSlaveError) in(PSLVERROR)
+  }
+}
+// end Apb3
+
+case class Apb3User(apbConfig: Apb3Config) extends Component {
```
numero-744, line 33:
```suggestion
// start usage
case class Apb3User(apbConfig: Apb3Config) extends Component {
```

source/SpinalHDL/Examples/Simple ones/apb3.rst
```diff
@@ -61,55 +61,24 @@ Implementation
...
-     val apb = slave(Apb3(apbConfig))
-   }
-
-   io.apb.PREADY := True
-   when(io.apb.PSEL(0) && io.apb.PENABLE){
-     //...
-   }
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/simple/Apb3.scala
+   :language: scala
+   :start-at: case class Apb3User(
```
numero-744, line 83:
```suggestion
   :start-after: // start usage
```

numero-744 (commented):
It is about 2:30 am but it was soooo interesting! Thanks for this beautiful PR!

* Maybe the path of the examples could be simplified? For instance flattening `spinaldoc/example` as just `spinaldoc`?
* Maybe the path should be `test` instead of `main`, to enable unit-testing beyond compilation/syntax?
* In the suggestions I suggest replacing ``` :code:`…` ``` by ``` ``…`` ```
* The goal of my suggestions is to have a more reproducible / uniform method to add beacons, which depends less on the current state of the code (for instance the fact that no functions have `{}`) to avoid issues on the long term.
* I will have to re-check for `(` after each `:start-at:` (when applicable)
* Do we need all these `import scala.language.postfixOps`?

About the method to put beacons:

* Adding `(` after `:start-at:`: good catch, but what if there is not? For instance `class Xxx extends Component`, do we put `extends` at the end?
* For `// end Item` there is the same issue, maybe we should  add a `.` at the end, but this way `// end Item io` should have a space as a separator, not a `.`, what do you think about it?
* For `// Item convenience functions` beacon, maybe the shorter `// Item helpers` would be enough?
* In the last commits the use was more complex, harder to make uniform, mainly because full examples were added at the end. Is there a way to put a link to the full source file, instead?

numero-744:
Another way to add beacons could be to give a name to each sample to include and use only this name:

> Useful cases of these option is working with tag comments. :start-after: [initialized] and :end-before: [initialized] options keep lines between comments:
>
> ```python
> if __name__ == "__main__":
>     # [initialize]
>     app.start(":8000")
>     # [initialize]
> ```

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

andreasWallner:
>     * Maybe the path of the examples could be simplified? For instance flattening `spinaldoc/example` as just ~`spinaldoc`~ `example`? EDIT: I've just understood that `example` comes from the chapter name. On the other hand do we need the `spinaldoc` part?

I don´t think so: the spinaldoc part is the name of the scala library being built, the rest I wanted to match the RTD structure so that stuff can be found more easily.

>     * Maybe the path should be `test` instead of `main`, to enable unit-testing beyond compilation/syntax?

With the next PR I'll add a test folder so that we can get some automation going. (Also automatic generation of everything, that's much harder to do w/o scalatest.

>     * I will have to re-check for `(` after each `:start-at:` (when applicable)
> 
>     * Do we need all these `import scala.language.postfixOps`?

Depends: the code currently compiles w/o, but checking here: https://www.scala-lang.org/api/2.13.3/scala/language$.html and here: https://docs.google.com/document/d/1nlkvpoIRkx7at1qJEZafJwthZ3GeIklTFhqmXMvTX9Q/edit this is at least a warning, and may become an error in upcoming versions of scala. IntelliJ and others already warn for it, that's why it's in there. But we can remove it, I'd keep it though. Some uses of the postfix notation are by now deprecated, see e.g. here: https://docs.scala-lang.org/style/method-invocation.html#higher-order-functions

(postfixOps are e.g. our `bits` or `Hz` operators)

> About the method to put beacons:
> 
>     * Adding `(` after `:start-at:`: good catch, but what if there is not? For instance `class Xxx extends Component`, do we put `extends` at the end?

We could do that, also often the classes will have parameters making this work. (apart from that we should prefer `case class` anyways - we even have that somewhere in the RTD).
On the other hand there is no need to overdo this: if someone changes an example they have to look at the render anyhow and act accordingly, I don't see a need for strict rules here.

>     * For `// end Item` there is the same issue, maybe we should  add a `.` at the end, but this way `// end Item io` should have a space as a separator, not a `.`, what do you think about it?

I did not really encounter that as an issue since the end marker is searched starting from the start marker, that makes it quite unlikely to hit such error cases.

>     * For `// Item convenience functions` beacon, maybe the shorter `// Item helpers` would be enough?

I guess either works, not overly concerned there.

>     * In the last commits the use was more complex, harder to make uniform, mainly because full examples were added at the end. Is there a way to put a link to the full source file, instead?

Since I kept the example pages as they were this made sense. Even if we just linked to the file I'd still try to got with the current setup - you'd have the same stuff in there regardless whether you link the file or include it.
In general I'd have liked to be able to link to full sources (also easier for people playing around and needing the full imports, etc.), but in a quick search I could only find https://www.sphinx-doc.org/en/master/usage/extensions/viewcode.html which seemed to be limited to Python code. I have written one or the other sphinx extension, but currently I don't really want to add to my TODO list ;-)

examples/src/main/scala/spinaldoc/examples/advanced/JTAG.scala
```diff
@@ -177,7 +177,7 @@ class JtagInstructionIdcode[T <: Data](value: Bits)(tap: JtagTapAccess, instruct
 }
 // end idcode
 
-class SimpleJtagTap extends Component {
+case class SimpleJtagTap() extends Component {
```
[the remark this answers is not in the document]
andreasWallner: Not sure that's a good reason: to me this screams for mixups, especially if the parameters have the same type (consider that you'd have to spot the difference between `new MyComponent(bits)` and `MyComponent(bits)` - any might not yet have seen the companion object). I'd rather use a differently named function in the companion object anyways.

andreasWallner:
Regarding whole code files: we could just link to github...

numero-744:
> apart from that we should prefer case class anyways - we even have that somewhere in the RTD

I remember there is this sentence in RTD but IIRC most components are just `class` so that is not logical. I think your update adding `case` is good so that the guideline matches the examples, and we should continue this way (we should update SpinalTemplateSbt, too, I'll do that later™).

Else I'm okay with your last comment.

examples/src/main/scala/spinaldoc/examples/advanced/JTAG.scala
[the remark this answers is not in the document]
numero-744: For the companion object with `apply(input: InputType): OutputType` I think we'll be able to discuss it when I PR this idiom suggestion (after #131 is completed).

numero-744:
Yes I was thinking to linking to GitHub but I'm afraid of versions mismatches, if the code changes…

numero-744:
XD

andreasWallner:
> I remember there is this sentence in RTD but IIRC most components are just `class` so that is not logical. I think your update adding `case` is good so that the guideline matches the examples, and we should continue this way (we should update SpinalTemplateSbt, too, I'll do that later™).

I'll repost my reply to one of your suggestions above:

Not sure that's a good reason: to me this screams for mixups, especially if the parameters have the same type (consider that you'd have to spot the difference between new MyComponent(bits) and MyComponent(bits) - any might not yet have seen the companion object). I'd rather use a differently named function in the companion object anyways.

numero-744:
It is here: https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Getting%20Started/Scala%20Guide/coding_conventions.html#class-vs-case-class

*A few lines later*

https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Getting%20Started/Scala%20Guide/coding_conventions.html#case-class

```scala
class Fifo extends Component {

}

class Counter extends Area {

}

case class Color extends Bundle {

}
```

:rofl:

andreasWallner:
I went through the files once again and had a look that the markers are more uniform - didn't use the suggestion from the interface though as it was easier to do it with two side-by-side windows (and check by rebuilding). So they might be slightly different.

andreasWallner:
Ah - I made a mistake when rebasing, give me a second to fix that... didn't want to change the old commits

numero-744:
I should have opened a PR targetting your branch, would have been simpler than GH suggestions XD

Ah I use `git rebase -i` a lot, I often use the `f` flag to squash commits.

I don't know if GH will be able to find back the old commits… We'll see :crossed_fingers:

andreasWallner:
So...this is much better^^

numero-744:
FYI I'm adding this `case class` for the template in a PR https://github.com/SpinalHDL/SpinalTemplateSbt/pull/21/commits/d40940a924cdf7be40b48cffc528eb2716cde9ad

source/SpinalHDL/Examples/Advanced ones/jtag.rst
```diff
@@ -152,177 +86,85 @@ First we need to define how an instruction could interact with the JTAG TAP core
...
-   }
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/advanced/JTAG.scala
+   :language: scala
+   :start-at: class JtagInstruction
+   :end-before: // end instruction
 
 .. note::
    | About the Component.current.addPrePopTask(...) : 
-   | This  allow you to call the given code at the end of the current component construction. Because of object oriented nature of JtagInstruction, doCapture, doShift, doUpdate and doReset should not be called before children classes construction (because children classes will use it as a callback to do some logic)
+   | This allows you to call the given code at the end of the current component construction. Because of object oriented nature of JtagInstruction, doCapture, doShift, doUpdate and doReset should not be called before children classes construction (because children classes will use it as a callback to do some logic)
```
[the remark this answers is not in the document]
numero-744: Up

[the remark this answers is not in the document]
numero-744: Up

source/SpinalHDL/Examples/Intermediates ones/vga.rst
```diff
@@ -214,38 +149,13 @@ The logic that generates horizontal and vertical synchronization signals is quit
...
-       }
-     }
-     val h = HVArea(io.timings.h, True)
-     val v = HVArea(io.timings.v, h.syncEnd)
-   }
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/intermediate/VGA.scala
+   :language: scala
+   :start-after: // end VgaCtrl io
+   :end-before: // end VgaCtrl HVArea
+   :prepend: case class VgaCtrl(rgbConfig: RgbConfig, timingsWidth: Int = 12) extends Component {
```
numero-744, line 156: Ooh, really nice!

source/SpinalHDL/Examples/Simple ones/pll_resetctrl.rst
```diff
@@ -6,31 +6,24 @@ Let's imagine you want to define a ``TopLevel`` component which instantiates a P
...
 The following imports will be used in code examples on this page:
 
-.. code-block:: scala
-
-   import spinal.core._
-   import spinal.lib._
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/simple/PLL.scala
+   :language: scala
+   :start-at: import spinal.core._
+   :end-at: import spinal.lib._
```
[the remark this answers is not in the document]
numero-744: I changed my mind XD

numero-744:
Okay so I think there is only one missing dot here: https://github.com/SpinalHDL/SpinalDoc-RTD/pull/148/files#diff-1986a0a5a9b8860eee568fe2e8c5b7fa1a5c9d3d58dd0e40cc978b34fe57990bR113

source/SpinalHDL/Examples/Intermediates ones/vga.rst
```diff
@@ -192,20 +136,11 @@ Component and io definition
...
-       val frameStart = out Bool
-       val vga = master(Vga(rgbConfig))
-     }
-     ...
-   }
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/intermediate/VGA.scala
+   :language: scala
+   :start-at: case class VgaCtrl(
+   :end-before: // end VgaCtrl io
+   :append: ...
```
numero-744, line 143: Hmm indentation is missing here?

source/SpinalHDL/Examples/Intermediates ones/vga.rst
```diff
@@ -214,38 +149,13 @@ The logic that generates horizontal and vertical synchronization signals is quit
...
-     val h = HVArea(io.timings.h, True)
-     val v = HVArea(io.timings.v, h.syncEnd)
-   }
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/intermediate/VGA.scala
+   :language: scala
+   :start-after: // end VgaCtrl io
+   :end-before: // end VgaCtrl HVArea
+   :prepend: case class VgaCtrl(rgbConfig: RgbConfig, timingsWidth: Int = 12) extends Component {
+             ...
+   :append: ...
```
numero-744, line 158: Hmm indentation is missing here?

source/SpinalHDL/Examples/Intermediates ones/vga.rst
```diff
@@ -254,26 +164,13 @@ Interconnections
...
-     io.vga.colorEn := colorEn
-     io.vga.color := io.pixels.payload
-   }
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/intermediate/VGA.scala
+   :language: scala
+   :start-after: // end VgaCtrl HVArea
+   :end-before: // end VgaCtrl connections
+   :prepend: case class VgaCtrl(rgbConfig: RgbConfig, timingsWidth: Int = 12) extends Component {
+             ...
+   :append: ...
```
numero-744, line 173: Hmm indentation is missing here?

source/SpinalHDL/Examples/Simple ones/sinus_rom.rst
```diff
@@ -35,66 +32,38 @@ Let's imagine that you want to generate a sine wave and also have a filtered ver
...
-     //Function used to generate the rom (later)
-     def sinTable = for(sampleIndex <- 0 until sampleCount) yield {
-       val sinValue = Math.sin(2 * Math.PI * sampleIndex / sampleCount)
-       S((sinValue * ((1<<resolutionWidth)/2-1)).toInt,resolutionWidth bits)
-     }
+.. literalinclude:: /../examples/src/main/scala/spinaldoc/examples/simple/SineRom.scala
+   :language: scala
+   :start-at: case class SineRom(
+   :end-at: }
+   :append: ...
```
numero-744, line 47: Hmm indentation is missing here?

numero-744 (commented):
I'm checking the resulting HTML.

I don't know if there is a possible fix for that

source/SpinalHDL/Examples/Intermediates ones/vga.rst
numero-744: Ah, same here, indentation missing for the `...`

source/SpinalHDL/Examples/Intermediates ones/vga.rst
andreasWallner: You mean the indentation of the ellipsis? I tried a few thing, but none worked:
Using a literal text like
```
:append: `  ...`
```
get taken literally and introduces ` into the literal block

Defining a replacement via `.. |nbsp| unicode:: 0xA0 ` also gets taken literally. Looking at the code of literalinclude I'm not sure there is a way to do that since the text in append/prepend is not parsed.

numero-744: So I guess we can leave this as-is? Not a big issue I think

numero-744:
@andreasWallner LGTM!

About your "What remains WIP:" in the description:

* have you noticed https://github.com/SpinalHDL/SpinalDoc-RTD/pull/148#issuecomment-1317757865
* do you want to keep that for another PR?

If so I merge this PR

andreasWallner:
I'd get this landed and continue from there - for most of the open stuff we have tasks already anyways. I'm currently putting together the PR to build this with scalatest, scalafmt should be easy going from there.

numero-744:
I don't understand XD "get this landed" means we should merge it right now, or we should do the last few things then merge it really soon?

andreasWallner:
yes, I meant to merge it now XD

PR #153, opened 2022-11-20
Add graySequential enum encoding

andreasWallner:
Add encoding added in https://github.com/SpinalHDL/SpinalHDL/pull/967
Also fixes broken syntax highlighting on the FSM page.

For now I did not extract the code examples to our checked build, I'd leave that for a separate commit that does not add any now code to make the review easier.

### DIFF
--- a/source/SpinalHDL/Libraries/fsm.rst
+++ b/source/SpinalHDL/Libraries/fsm.rst
@@ -140,6 +140,36 @@ Transitions
 These two functions can be used inside state definitions (see below) or using ``always { yourStatements }``,
 which always applies ``yourStatements``, with a priority over states.
 
+State encoding
+^^^^^^^^^^^^^^
+
+By default the FSM state vector will be encoded using the native encoding of the language/tools the RTL is generated for (Verilog or VHDL).
+This default can be overriden by using the ``setEncoding(...)`` method which either takes a ``SpinalEnumEncoding`` or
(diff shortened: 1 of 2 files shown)


source/SpinalHDL/Libraries/fsm.rst
```diff
@@ -140,6 +140,36 @@ Transitions
...
+   }
+
+.. code-block:: scala
+   :caption: Using a custom encoding
+
+   val fsm = new StateMachine {
+     val stateA = new State with EntryPoint
+     val stateB = new State
+     ...
+     setEncoding((stateA, 0x23), (stateB, 0x22))
```
numero-744, line 166:
```suggestion
     setEncoding((stateA -> 0x23), (stateB -> 0x22))
```

I prefer this syntax to define an encoding

source/SpinalHDL/Libraries/fsm.rst
```diff
@@ -140,6 +140,36 @@ Transitions
...
+   }
+
+.. warning:: When using the ``graySequential`` enum encoding no check is done to verify that the FSM transitions only produce
```
numero-744, line 169:
```suggestion
.. warning:: When using the ``graySequential`` enum encoding, no check is done to verify that the FSM transitions only produce
```

I think putting a `,` here helps to see the "no" by slowing down reading there, else reading too fast I skip this little but so important word XD

numero-744:
Thanks!

PR #162, opened 2022-12-07
getting started/sbt: compile -> Test/compile

numero-744:
Tests are being moved to specific folder so this line should be updated.

Note: this is an untested quickfix from GH web interface

### DIFF
--- a/source/SpinalHDL/Getting Started/SBT.rst
+++ b/source/SpinalHDL/Getting Started/SBT.rst
@@ -24,7 +24,7 @@ fetch dependencies so it can take time the first time:
 
 .. code-block::
 
-   compile
+   Test/compile
 
 Actually you never need to just ``compile`` as it is done automatically when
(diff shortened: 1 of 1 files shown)


numero-744:
Hmm this PR is good enough: if we move simulations to `tb` we have to do `Test/runMain` so it would need more explanations :thinking:

Dolu1990:
> I think runMain and test could not be mixed. runMain is used to generate the RTL for the design, while test is used to test the design.
so either use sbt runMain xxxx.xxxx or sbt testOnly xxxx.xxxx for a simple case.

I would say, the main distinction between main / Test is that main will be contained in the release jar, while test will not be included.
The scala test framework is something else (coupled to Test)

numero-744:
There it also that you need Test/runMain to run an App in the test (tb) folder, and just runMain if it is in the main (hw) folder

Le 8 décembre 2022 21:48:32 UTC, Dolu1990 ***@***.***> a écrit :

PR #165, opened 2022-12-18
Add Elitestek to companies using SpinalHDL page

Eric1993-mkt:
Elitestek has used the VexRISC-V core in FPGAs and applied in multi applications in worldwide customers.


Readon:
I'm sorry for changing the code directly without confirmation. Just added a space bar after the semicolon.
If you also have an English version page, an update is also welcome.
Thanks, ^^

PR #169, opened 2023-01-03
update verilator's version from v4.040 -> v4.218

hanm2019:
in [tool.sh](https://github.com/SpinalHDL/SpinalHDL/blob/4a21ab4cca95b35fbb12189077e7609e7b40dbb9/tools.sh#L9), the version of verilator is updated to `v4.218`, while the Doc still recommands the `v4.040`, maybe we need to update the Doc


numero-744:
Thanks!

PR #176, opened 2023-03-05
Flow simulation support and compile example

andreasWallner:
Add simulation support section like in "Stream" to "Flow" and move Flow example to examples to be compiled.


Dolu1990:
Thanks :D

PR #177, opened 2023-03-05
Fix order of slices from subdivideIn

andreasWallner:
It strikes me as weird that no one noticed this but
```scala
case class TestX() extends Component {
  val i = in Bits(128 bit)
  val o = out Bits(32 bit)
  o := i.subdivideIn(32 bit)(0)
}
object TestXX extends App {
  SpinalVerilog(TestX())
}
```
generates
```
module TestX (
  input      [127:0]  i,
  output     [31:0]   o
);


  assign o = i[31 : 0];

endmodule
```
Which indicates that the ordering was documented incorrectly

### DIFF
--- a/source/SpinalHDL/Data types/Int.rst
+++ b/source/SpinalHDL/Data types/Int.rst
@@ -458,10 +458,10 @@ Misc
    // Subdivide
    val sel = UInt(2 bits)
    val mySIntWord = mySInt_128bits.subdivideIn(32 bits)(sel)
-       // sel = 0 => mySIntWord = mySInt_128bits(127 downto 96)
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hoo yes right :D
Thanks

I'm currently in travel, on my laptop which has a veeeeeryyy old version of web browser installed, I can't even click on "merge pull request" the thing is too bugy XD

Feel free to merge :)

andreasWallner:
@Dolu1990 I don't have a merge button, but thanks @wifasoi for merging^^

Dolu1990:
@andreasWallner

Yeah XD
I checked to add you the rights, but that too is bugy on my old linux install XD

PR #179, opened 2023-05-08
Cleanup dev branch

dlmiles:
* makes Miscellaneous content accessible (quite a bit of content, confirm it should be public)
* removed TravisCI SVG badge that 404s
* Fix error/warning due to missing column in table "Range"
* LICENSE file missing first character

These 2 items don't necessarily need to be in PR, but are required in my source branch so that it can be built in a github action (maybe I can rename the source branch of this PR from 'https://github.com/dlmiles/SpinalDoc-RTD/tree/dev' to 'https://github.com/dlmiles/SpinalDoc-RTD/tree/dev-pr179' to if separation is required:
* bin/convert-wrapper
* .github/workflows/build-and-depoy.yml  this is my all in one action test/live linkcheck/no-linkcheck  single/multi  deploy

EDIT this build failure https://github.com/SpinalHDL/SpinalDoc-RTD/actions/runs/4928725690/jobs/8807447131 is the reason the bin/convert-wrapper commit exists.  Although only 1of3 is included here due to differences between dev and master at this time.

The current actions seem to be using older version of things that may not be getting maintained going forwards:
* The docker image for Sphinx only goes to 5.2.0 (which is fine for now, but they appear to have stopping updating that image now https://hub.docker.com/r/sphinxdoc/sphinx-latexpdf/tags , maybe they have an alternative, Sphinx 7.x is the latest, although 5.3.0 is only 6 months old at this time)
* The gh-pages branch management action JamesIves/github-pages-deploy-action@4.1.4 (maybe switch to 'dev' branch?), this maybe using using NodeJS 12 internally that gh-actions will make extinct
* This action appears to 'chmod +rw' content internally, this seems a bad choice for online information even if hosted by github managing security.
* The method appear to commit to the source repository the CI build is coming from, this is not a method/pattern I like to see in use myself, a CI is a pipeline with data flowing in one direction, never back to its own source, accidents happen.  So having another repo with the sole purpose to keep a copy of the final HTML makes sense, but not a branch in the same repo being driven by automation.  Just my opinion on this, so this is not something (a github write token) I authorize even in my fork to work on PR, so this makes it untested in any of my commits.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-reative Commons Legal Code
+Creative Commons Legal Code
 
 CC0 1.0 Universal
 
--- a/source/SpinalHDL/Data types/bits.rst
+++ b/source/SpinalHDL/Data types/bits.rst
@@ -274,6 +274,7 @@ Misc
      - Int
    * - x.bitsRange
      - Return the range (x.high downto 0)
+     - Range
    * - x.valueRange
      - Return the range (x.minValue downto x.maxValue). Note can't be used for value which overflow the JVM Int capacity.
      - Range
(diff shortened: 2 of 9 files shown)


dlmiles:
3ac0e94c3 is the final version on this PR this can be applied now, then when that passes CI.

Currently the site is `https://spinalhdl.github.io/SpinalHDL/` HTTP/404  and the `make linkcheck` will fail ?  preventing from publishing from CI ?

Dolu1990:
Thanks ^^

PR #181, opened 2023-05-28
describe how wrap method work for UInt type.

Readon:
describe SpinalHDL/SpinalHDL#675.

### DIFF
--- a/source/SpinalHDL/Data types/Int.rst
+++ b/source/SpinalHDL/Data types/Int.rst
@@ -238,8 +238,8 @@ Arithmetic
 
 .. note::
 
-   Notice how simulation assertions are made here (with ``===``), as opposed to elaboration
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^

Are wrap operations similar to SInt ones ? else, what differs ? I'm trying to understand the application ^^

Readon:
> 

What do you mean wrap function of ```SInt```? can you give me an example or link?

For example, in TCP protocol the ACK number is used to acknowledge the sent packets with SEQ number smaller than it.  
However, this number is in ```UInt``` type and always increasing. It would wrap around to 0 while overflow, so that it has to be compared with wraping capability.

dlmiles:
@Readon Rebase onto `master` ?   Master is 156 ahead of `dev`.  Is there a way to do that without closing PR and re-opening against `master` ?

PR #183, opened 2023-05-31
describe how wrap method work for UInt type.

Readon:
A fix of #181 and switch to master branch.

### DIFF
--- a/source/SpinalHDL/Data types/Int.rst
+++ b/source/SpinalHDL/Data types/Int.rst
@@ -284,8 +284,8 @@ Comparison
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
So if i understand well those comparison operators will look at the smalest interval between the two values to figure out the relative position between the two operands ?

Dolu1990:
> So if i understand well those comparison operators will look at the smalest interval between the two values to figure out the relative position between the two operands ?
yes

I think the doc would need a description of that sort, as currently, "only" a single example is provided (0xFF-0x01) which may not be enough to generalize

PR #187, opened 2023-06-04
Tranche 2

dlmiles:


source/SpinalHDL/Data types/bits.rst
```diff
@@ -282,11 +307,15 @@ Misc
      - Return the index of the MSB (highest allowed index for Bits)
      - Int
    * - x.msb
-     - Return the most significant bit
+     - Return the most significant bit value
```
andreasWallner, line 310: I find this one to be more confusing with the change, "most significant bit" is a well defined term, if you add "value" I get suspicious...

source/SpinalHDL/Data types/bits.rst
```diff
@@ -282,11 +307,15 @@ Misc
...
      - Bool
    * - x.lsb
-     - Return the least significant bit
+     - Return the least significant bit value
      - Bool
+   * - x.reverse
+     - Return a representation of the same instance in reverse bit order,
```
andreasWallner, line 316:
```suggestion
     - Return a copy of x with reversed bit order
```
Use the same wording as below?

source/SpinalHDL/Data types/bits.rst
```diff
@@ -297,18 +326,29 @@ Misc
      - Subdivide x in multiple slices of y bits, y: Int
      - Vec(Bits, w(x)/y)
    * - x.resize(y)
-     - Return a resized copy of x, if enlarged, it is filled with zero, y: Int
+     - Return a resized copy of x, if enlarged, it is filled with zero
```
andreasWallner, line 329:
```suggestion
     - Return a resized copy of x, if enlarged, it is extended with zero
```

source/SpinalHDL/Data types/bool.rst
```diff
@@ -33,14 +33,16 @@ The syntax to declare a boolean value is as follows: (everything between [] is o
      - Create a Bool assigned with ``false``
      - Bool
    * - Bool(value: Boolean)
-     - Create a Bool assigned with a Scala Boolean(true, false)
+     - Create a Bool assigned with a value from a Scala Boolean type (true,
+       false).  This implicitly converts to ``True`` or ``False`` as
```
andreasWallner, line 37: It does not implicitly convert since it's a Bool all along?

source/SpinalHDL/Data types/bool.rst
```diff
@@ -33,14 +33,16 @@ The syntax to declare a boolean value is as follows: (everything between [] is o
...
+       necessary during elboration phase of HDL generation.
      - Bool
 
 
 .. code-block:: scala
 
-   val myBool_1 = Bool()          // Create a Bool
-   myBool_1 := False            // := is the assignment operator
+   val myBool_1 = Bool()        // Create a Bool
+   myBool_1 := False            // := is the assignment operator (like verilog <=)
```
andreasWallner, line 45: If we want to add that information we should make a "Help for Verilog people" chapter rather than this one lone mention?

source/SpinalHDL/Data types/bool.rst
```diff
@@ -115,13 +124,21 @@ Logic
...
 
-  // mind the order of assignments!
+  // mind the order of assignments!  last one wins
   val g = RegInit(False) setWhen(req) fallWhen(ack)
   // equivalent to g := ((!g) && req) || (g && !ack)
 
 Edge detection
 ~~~~~~~~~~~~~~
 
+This feature will always imply a :ref:`RegNext <regnext>` pattern to track the state history
```
andreasWallner, line 134:
```suggestion
All edge detection functions will instantiate a register via :ref:`RegNext <regnext>` to get a delayed value of the ``Bool`` in question.
```
?

source/SpinalHDL/Data types/bundle.rst
```diff
@@ -80,6 +80,8 @@ Comparison
    color2.b := 0
 
    myBool := color1 === color2
```
andreasWallner, line 82: I'd rather add a comment "// compares all members" like above with Vec

andreasWallner (changes requested):

source/SpinalHDL/Introduction/A simple example.rst
```diff
@@ -102,8 +102,8 @@ Finally, there is the component logic:
      io.flag := (counter === 0) | io.cond1
 
 ``counter`` is a register containing an 8-bits unsigned integer, with the
-initial value 0. Assignments to a registers can be read only after the next
-clock sampling.
+initial value 0. Assignments to change the state of a register can be read only
```
dlmiles, line 105: initial value 0. Assignments to change the state of a register are available for read-back only

source/SpinalHDL/Data types/Int.rst
```diff
@@ -206,7 +208,7 @@ Arithmetic
      - Subtraction with carry
      - T(max(w(x), w(y)) + 1 bits)
    * - x -| y
-     - Subtraction by sat carry bit
+     - Subtraction by saturation carry bit
```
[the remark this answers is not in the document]
dlmiles: Can this be both MIN and MAX ?  or does this always saturate to MIN ?   If so, what is that based on the +ve or -ve of the addend ?  Does this mean the same is true for the Addition ?  Trying understand the difference in description between ADD and SUBTRACT.

source/SpinalHDL/Data types/Int.rst
```diff
@@ -308,6 +316,9 @@ Type cast
    * - x.asBools
      - Cast into a array of Bool
      - Vec(Bool(), w(x))
+   * - x.asBool
+     - Cast LSB into a signal bit returned as Bool
```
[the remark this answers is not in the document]
dlmiles: I think that should have been `... a single bit ...`

source/SpinalHDL/Data types/Int.rst
```diff
@@ -447,10 +458,17 @@ Misc
    * - mySInt.absWithSym
      - Return the absolute value of the UInt value with symmetric, shrink 1 bit
      - UInt(w(mySInt) - 1 bits)
+   * - x.getZero
+     - Return a new instance of type T that is assigned a constant value of zeros the same width as x.
+     - T(0, w(x) bits)
```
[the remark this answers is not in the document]
dlmiles: Yes but.... `T(0, w(x) bits)` is a ZERO number already.
I could not do `T(~0, w(x) bits)` reliably, but to the potential Int/Long limits in Scala, so the only way I could think of expressing the result was to `T(0, w(x) bits).setAll()` as that works for 4096 bit wide value.
I am +0 and not fussed either way.

source/SpinalHDL/Data types/bool.rst
dlmiles: This is obviously an Explicit conversion from Boolean to Bool not sure what I was thinking that day.  Maybe if we replace the word `implicit` with `explicit` does it make sense.

It starts off as a Boolean and becomes a Bool, like a regular JVM language type conversion.

source/SpinalHDL/Data types/bool.rst
dlmiles: `...instantiate an additional register via ...` ?  to clarify the register being talked about ?   It may sound verbose to us to restate that but I ask myself how the documentation reads after it has been put through an automated translator into another natural language.

source/SpinalHDL/Data types/Int.rst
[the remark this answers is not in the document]
andreasWallner: Didn't think of the case that a negative value is added to an SInt, so addition can also be either min or max (but only for SInt). Instead of overflowing it saturates -> I'd vote for keeping the simple wording and maybe add a link to https://en.wikipedia.org/wiki/Saturation_arithmetic

source/SpinalHDL/Data types/Int.rst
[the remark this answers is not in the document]
andreasWallner: Why, LSB as in "least significant bit" is quite clear?

source/SpinalHDL/Data types/bool.rst
andreasWallner: Not sure google translate is a good test for this... but yes, let's go for "an additional register via"

source/SpinalHDL/Data types/bool.rst
andreasWallner: Do you think it would make more sense to describe this difference at the beginning of this chapter, a quick `Bool` vs. `Boolean`? In the sense of:
"The Bool type corresponds to a boolean value (True or False) or a single bit/wire in hardware. While named similarly it should not be confused with Scala `Boolean`s which don't describe hardware but truth values in the Scala generator code. To be used in hardware a `Boolean` must be converted to a `Bool`, i.e. using the below described `Bool` constructor."

dlmiles: Definitely :)

source/SpinalHDL/Data types/bits.rst
dlmiles: I see what you mean about the use of the word `copy` in the item below or so.  I would argue this is also a bad choice and the word `representation` is a better word there as well, relating to a `data representation` of the same underlying thing.  In reading the docs I was trying not to be too nit-picky over such things.  But if I were I would change that too and use a distinct language word less over-used than the word copy (can mean a lot of different things, so lots of room for misunderstanding).

source/SpinalHDL/Developers area/types.rst
```diff
@@ -772,7 +772,8 @@ Continuous Assignment Literals as signal declaration
 You can also use them in expressions to do three things at once :
 
 * Define a wire
-* Maintain the result of an equality operation with the constant value and another signal
+* Maintain the result of an equality operation in logic implementation with the constant
```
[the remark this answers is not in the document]
dlmiles: To explain adding the word `hardware` the question it is closing is:  are we talking about `Scala logic` of `Hardware logic` ?

source/SpinalHDL/Data types/bits.rst
```diff
@@ -92,29 +97,41 @@ Logic
    * - x.andR
      - AND all bits of x
      - Bool
-   * - x \>\> y
-     - Logical shift right, y: Int
+   * - | y = 1 // int
```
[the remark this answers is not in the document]
dlmiles: This is up there now... https://dlmiles.github.io/SpinalDoc-RTD/docs_multi/html/vTEST/SpinalHDL/Data%20types/bits.html#logic  apart from the extra whitespace towards bottom of cell I think it look better.

I have fixed the use of `int` instead of `Int` to draw attention to types involved.

source/SpinalHDL/Data types/Int.rst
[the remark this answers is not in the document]
dlmiles: Sorry IDD with the change.
I was correcting my original commit :)
I can't edit the reviewed code in github to make the change.  I have to force-push, which I can do once all the feedback has been worked through.  So maybe I +1 the change and will provide it when I make the push.
I don't want to push now as I don't want to mess up the review process until all matters seem resolved/have solution.

source/SpinalHDL/Data types/Int.rst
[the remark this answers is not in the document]
dlmiles: I don't mind the extra info, just asking the obvious question when someone reads it.

But if I were to keep it simple (I assume you mean by that a summary description of operation, without the minValue/maxValue details of the implementation) :

`Subtraction of subtrahend with ::ref:<saturation https://en.wikipedia.org/....>`.

The 'carry bit' is not a thing ?   The 2 words 'carry bit' seem out of place to me, if there is no extra 'carry bit' information that is an input or output.

source/SpinalHDL/Data types/bits.rst
```diff
@@ -92,29 +97,41 @@ Logic
...
+       | x \>\> y
+     - | Logical shift right, y: Int
+       | Result may reduce width
```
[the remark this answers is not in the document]
dlmiles: This is up there now... https://dlmiles.github.io/SpinalDoc-RTD/docs_multi/html/vTEST/SpinalHDL/Data%20types/bits.html#logic apart from the extra whitespace towards bottom of cell I think it look better.

The commit you are looking at looks like contents of the page up there now.

The Vertical Space is a CSS matter I can probably resolve later, not really something I would block the new content / commit on.

dlmiles (commented):
Oh there is a button "Finish your Review" lets see what that does :)

source/SpinalHDL/Data types/Int.rst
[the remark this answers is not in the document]
andreasWallner: I guess the carry bit was mentioned because the HW generated does the arithmetic operation with operands extended and uses the carry bit to mux between the result or the max value.

source/SpinalHDL/Data types/Vec.rst
```diff
@@ -85,6 +85,8 @@ Comparison
    val vec1 = Vec(SInt(8 bits), 2)
 
    myBool := vec2 === vec1  // Compare all elements
+   // is equivalent to:
```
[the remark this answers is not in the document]
andreasWallner: Yes I find it possibly confusing because the equivalence is only for the specific case of Vecs with 2 elements.
Maybe we make it an "in this case equivalent..."?
No need to uncomment it, if you want it to be compilable we could just make it a `val otherBool = ...` - but in the end many of these examples show single lines of usage and would not be very useful taken together... (which was one of the things that stopped me when I tried to move them to the compiled project - much to rewrite and IMO little gain in trying to find the n-th way of saying "someBool"...)

source/SpinalHDL/Data types/bits.rst
```diff
@@ -41,7 +41,12 @@ The syntax to declare a bit vector is as follows: (everything between [] is opti
 .. code-block:: scala
 
    // Declaration
-   val myBits  = Bits()     // the size is inferred     
+   val myBits  = Bits()     // the size is inferred through later assignment
```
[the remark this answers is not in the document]
andreasWallner: I should have been clearer: The width is inferred from the assignments (actually the biggest known width) - so 6 bit width is inferred in your example.
But you try to assign a 4 bit value to a (inferred size) 6 bit value, which generates an error.

source/SpinalHDL/Data types/bits.rst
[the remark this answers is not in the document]
andreasWallner: While the newlines are now fixed, to some other things remain:
- The table entries now take up 3 lines (2 lines of content, one empty line)
- "int" is not a type, the correct scala type is "Int"
- I still think it makes it less readable to restate all information twice, both the type of "y" and the size of the output result are present

source/SpinalHDL/Data types/bits.rst
andreasWallner: Interesing: to me, if you ask for the "the most significant bit" you ask for the bit, not it's position. - probably an indication that we should make this clear.

I'm a bit rusty on my english and not sure if that should then be "the most significant bit value" or "the most significant bits value", maybe go for "the value of the most significant bit"?

source/SpinalHDL/Data types/bits.rst
andreasWallner: I had a quick look at the code: it looks to me like this should actually be `reversed` - overlooked that initially
https://github.com/SpinalHDL/SpinalHDL/blob/af55caaf3e9821eed2b2ad2d5bfda742716bcb43/core/src/main/scala/spinal/core/Bits.scala#L237
I used copy since the reversed vector will actually not let you modify the original one, i.e.
```scala
val x = Bits(5 bit)
val y = x.reversed
when(x(0)) { y(3) := True }
```
generates
```verilog
y = {x[0],{x[1],{x[2],{x[3],{x[4]}}}}}
if(...) begin
  y[3] = 1'b1;
end
```

source/SpinalHDL/Data types/bool.rst
andreasWallner: I understand what you mean. But is this in the end a helpful equivalence? `:=` is used in places were verilog would use `<=` or `=`? But if you think it helps that much we can also keep it.

And no, I dont think we need "equal representation" ;-) VHDL equivalents have their own page where we have to space to actually mention the details.

dlmiles: As someone still in their first year of using SpinalHDL, I would have found it useful to get a frame of reference.

With regards to `<=` and `=` differences, it does not matter for the purpose of frame of refrence, the most often used type in verilog assignment type is `<=` by maybe 99:1 in the verilog I have seen.  YMMV

Also the snippet in this code example box is really targeting a novice/beginner to intermediate user of SpinalHDL.  I think an intermediate user and above will glance at that code example box and move on, on the basis there is nothing to learn in there.

dlmiles:
Everything is resolved at this time.  Please confirm/restate/point out anything missed.  Thanks

source/SpinalHDL/Data types/bits.rst
```diff
@@ -41,7 +42,15 @@ The syntax to declare a bit vector is as follows: (everything between [] is opti
...
 
    // Declaration
-   val myBits  = Bits()     // the size is inferred     
+   val myBits  = Bits()     // the size is inferred at first assignment
+   // ....
+   myBits := B("001010")    // Bits(6 bits) would be inferred
+   when(condxMaybe) {
+     myBits := B("110000")
+   } elsewhen(condxMaybeTwo) {
+     //myBits := B("00110000")	// Would Error: WIDTH MISMATCH (8 bits <- 6 bits)
```
andreasWallner, line 51:
```suggestion
     //myBits := B("0011")	// Would Error: we can't assign a 4 bit Bits
     to a 6 bit Bits w/o resizing
```
We should use your original example with 4 bits as it does not change which size is inferred - easier to understand IMO

source/SpinalHDL/Data types/bool.rst
```diff
@@ -9,7 +9,52 @@ Bool
 Description
 ^^^^^^^^^^^
 
-The ``Bool`` type corresponds to a boolean value (True or False).
```
andreasWallner, line 12: As mentioned in the chat - at least half of this should either go to the chapter on assignment or to to Introduction on Datatypes (and yes, we should fix the order that the Introduction comes first, but we can do that in a separate commit).
I'll still add some comments on the actual text.

source/SpinalHDL/Data types/bool.rst
```diff
@@ -9,7 +9,52 @@ Bool
...
+The ``Bool`` type corresponds to a boolean value (True or False) or a single bit/wire
+used in a hardware design.  While named similarly it should not be confused with
+Scala `Boolean` type which does not describe hardware but truth values in the Scala
+generator code.
+
+An important concept and rule-of-thumb to understand is that the Scala
+`Boolean` type is used in places where ahead-of-time HDL code-generation
```
andreasWallner, line 18:
```suggestion
`Boolean` type is used in places where elaboration-time HDL code-generation
```
The term "elaboration" is used in parts of the documentation and should ring a bell for people coming from other HW langs.

source/SpinalHDL/Data types/bool.rst
```diff
@@ -9,7 +9,52 @@ Bool
...
+decision making is occuring in Scala code.  Like any regular program it affects
```
andreasWallner, line 19: "It is a condition during generation of the hardware and does not become hardware."?
The current sentence is far from straightforward

source/SpinalHDL/Data types/bool.rst
```diff
@@ -9,7 +9,52 @@ Bool
...
+execution of the Scala program that is SpinalHDL at the time the program is
+being run to perform HDL code generation.
+
+Therefore the value of a Scala `Boolean` can not be observed from hardware,
+because it only exists ahead-of-time in the SpinalHDL program at the time of
+HDL code-gen.
+
+In scenarios where you might need this for your design, for example to pass a
```
andreasWallner, line 27: The general difference between scala types and SpinalHDL types would make sense in the introduction on datatypes.

source/SpinalHDL/Data types/bool.rst
```diff
@@ -9,7 +9,52 @@ Bool
...
+value (that maybe acting as a parameterized constant input) from Scala into your
+hardware design, you can type convert it to Bool with the constructor `Bool(value: Boolean)`.
+
+Similarly the value of a SpinalHDL `Bool` can not be seen at code-generation, all
+that can be seen and manipulated is the HDL construct concerning a `wire` and how it
+is routed (through modules/Components), driven (sourced) and connected (sunk).
+
+The signal direction of assignment operators `:=` is managed by SpinalHDL.
```
andreasWallner, line 35: are you trying to say that the right side of `:=` is assigned to the left side?

source/SpinalHDL/Data types/bool.rst
```diff
@@ -9,7 +9,52 @@ Bool
...
+(captures state) for a given assignment.
+
+Multiple uses of the assignment operator are allowed, such that it is normal
+for a signal wire to act as a source (provides a value to drive HDL state) to be
+able to connect and drive multiple inputs of other HDL constructs.  When a Bool
+instance used as a source the order the assignment statments appear or are
+executed in Scala does not matter, unlike when it is used as a sink
+(captures state).
+
+When multiple assignment operators drive the Bool (the Bool is on the
```
andreasWallner, line 47: Only if all but one of the assignments is in a condition, see chapter on assignments.

source/SpinalHDL/Data types/bits.rst
```diff
@@ -92,29 +101,41 @@ Logic
    * - x.andR
      - AND all bits of x
      - Bool
-   * - x \>\> y
-     - Logical shift right, y: Int
+   * - | y = 1 // Int
```
andreasWallner, line 104: I still think that it's easier to read without the duplicated information.

andreasWallner (commented):
In addition to these ones there is one comment still unresolved (ad `reverse`)

source/SpinalHDL/Data types/bool.rst
dlmiles: I am unable to relate your quoted sentence/question to the original text.

I do not understand your sentence in quotes, and do not think it is a better suggestion.  It lacks context, it doesn't explain what is a condition of what, it repeats the use of the hardware term in a bizarre reference to itself.

Unfortunately none of the concepts being conveyed are straightforward, it is just the nature of the complexity that exists and the elements of underlying knowledge needed to digest new concepts.  But we can try improve clarify.

source/SpinalHDL/Data types/bool.rst
dlmiles: I think you mean the `when` condition and other similar HDL constructs.  I found this to be a given and it has not been an important detail to understand at any point so far.

That is: Yes if I wrap my assignment in a Scala or hardware condition it is going to modulate if and when that assignment occurs.  This is a given.

If there is a specific contrived example that can highlight a not-obvious SpinalHDL precedence rule that breaks the general simple rule(s), maybe there should be a section demonstrating it, but I would think it would be under intermediate/advanced user areas (end of chapters/erratum) and not something brought up when explaining initial concepts.

Kind of learn this first... now you have understand that... in this scenario those rules break.  This is the exception of the rule.

source/SpinalHDL/Data types/bool.rst
dlmiles: Maybe it would,  I shall re-read the datatypes sections again, but not from a typo/grammar improvement stance, more of a novice/beginner understanding it stance.

I certainly did not find the Scala datatype/class diagram and other info useful until I first needed to use Scala generics on Data `[T <: Data]` in some code.  Only then did I found it super useful.  I did not in anyway find it useful to translate existing Verilog/HDL knowledge into SpinalHDL knowledge concerning data.

There are probably only around 5 major areas that are worth an explanation like this before there should be enough knowledge to understand reference material.  Bool is defiantly one place.  Bits/Ints is another (which is conceptually a `bus`),  Mux/when/conditions is another, Register is another, combinational expressions and signal direction another.

source/SpinalHDL/Data types/bits.rst
dlmiles: I think you are citing the TYPE info column as duplicate information ?  Maybe to you as an advanced user, but to a novice that is type information (like the heading says - I'm not suggesting a change here BTW).

It is much better to restate things as key-points in clear readable language than use a new foreign cryptic notation to the reader, that add additional load to the reader trying to decode/understand what is being read.

Maybe the right hand column would work well with a tooltip style hover over, explaining the meaning of the notation in plain language.

`The result is the same data type as the original at the same width.`

It would however remove it from the page for any PDF/screen-capture.

source/SpinalHDL/Data types/bool.rst
dlmiles: IDD, maybe the dual but similar meaning use of the term elaboration both
 * in SpinalHDL and 
 * also the next stage of the flow with Verilog/VHDL to achieve synthesis
Is the only thing not clear with this for a first time reader.

In online documentation maybe this can be a linked glossary term with hover-over which defines the term, to clarify the distinction to the first time reader and also provide reference to disambiguate.
This is more a matter for setting up such a glossary of terms/definitions to help here.

source/SpinalHDL/Data types/bool.rst
dlmiles: IMHO There are probably only around 5 major areas that are worth an explanation like this (a worked application of information to a specific domain) before the reader should have enough knowledge to understand reference style material.

Bool is defiantly one place.
Bits/Ints is another (which is conceptually a `bus`),
Mux/when/conditions is another,
Register is another
Combinational expressions, signals and signal direction another.

Maybe they would be improved in a tutorial/exercise form, with just a link/heading from the top of each relevant section with 'First time readers might with to work the tutorial on: Subject of tutorial' to gain a better understanding.

At the moment it is written in the form I wish I had seen to convey a number of concepts to think about at the same time, but my route was more examining the verilog output and working through the error messages :)

source/SpinalHDL/Data types/Vec.rst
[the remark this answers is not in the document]
dlmiles: I don't understand how it changes if there are 3 element or 100 elements.

To me it would be a "given" (that "in this case equivalent...").  I would expect the equivalent to relative directly to what is near-by and only for this example.  If I do something different may no longer hold true.

So the question is how much change is needed to get an unexpected result ?  That is not clear to me, can you think of an example ?

From the equivalent example the reader should be able to infer how more equality `===` operators are used and ANDed to reduce to a single Bool result.  Also how more element are compared for equality purposes.

If it helps we can use an additional 3 element example instead for clarify this point:

```
result := vec(0).a === vec(1).a && vec(1).a === vec(2).a && vec(2).a === vec(0).a &&
          vec(0).b === vec(1).b && vec(1).b === vec(2).b && vec(2).b === vec(0).b &&
          vec(0).c === vec(1).c && vec(1).c === vec(2).c && vec(2).c === vec(0).c
// although we can expect boolean algebra reduction to remove some of those comparisons when possible
//  AA === BB && BB === CC therefore CC === AA eliminating the need for that last comparison
```

This would be my first guess at how that works in a difference scenario.  Maybe SpinalHDL performs the boolean algebra reduction, maybe my hardware synthesis flow stage does, does not matter, unless we are trying to relate the example more to specifically how HDL will be generated ?  Functional equivalence (instead of absolute detail) is good enough for me in such an example.

So the best guess by the reader about a different scenario (than the example) is correct.

So how do you get an unexpected result that bites you in the ass ?

source/SpinalHDL/Data types/bits.rst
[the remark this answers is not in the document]
dlmiles: So we ended up as:
```
   // Declaration
   val myBits  = Bits()     // the size is inferred from the widest assignment
   // ....
   // .resized needed to prevent WIDTH MISMATCH error as natural width does not match inferred size
   myBits := B("1010").resized  // auto-widen Bits(4 bits) to Bits(6 bits)
   when(condxMaybe) {
     myBits := B("110000")      // Bits(6 bits) would be inferred for myBits, this is the widest assignment
   }
```

This does help clarify the inference of hardware matter from the inference of Scala matters, which is the usual use of the term.

source/SpinalHDL/Data types/Vec.rst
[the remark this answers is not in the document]
andreasWallner: Rereading it I agree, it should be clear from the context.

source/SpinalHDL/Data types/bool.rst
andreasWallner: I actually found it helpful in the beginning, it's very similar to diagrams that I was used to from VHDL...

Let's keep it this for now and I'll move it when I fix the order of the chapters.

andreasWallner:
Thanks for all the effort!

Dolu1990:
Thanks :D

PR #189, opened 2023-06-04
Data types/Int: updated

dlmiles:

https://dlmiles.github.io/SpinalDoc-RTD/docs_multi/html/vTEST/SpinalHDL/Data%20types/Int.html

See review marker looking for assistance to reword this area. Thanks

### DIFF
--- a/source/SpinalHDL/Data types/Int.rst
+++ b/source/SpinalHDL/Data types/Int.rst
@@ -103,13 +103,13 @@ Logic
      - Bitwise XOR
      - T(max(w(x), w(y)) bits)
    * - x.xorR
-     - XOR all bits of x
+     - XOR all bits of x (reduction operator)
      - Bool
    * - x.orR
-     - OR all bits of x
+     - OR all bits of x (reduction operator)
      - Bool
    * - x.andR
-     - AND all bits of x
+     - AND all bits of x (reduction operator)
      - Bool
    * - x \>\> y
      - Arithmetic shift right, y : Int
@@ -130,10 +130,12 @@ Logic
      - Logical shift left, y : Int/UInt
      - T(w(x) bits)
    * - x.rotateLeft(y)
-     - Logical left rotation, y : UInt/Int
+     - | Logical left rotation, y : UInt/Int
+       | The width of y is constrained to the width of log2Up(x) or less
      - T(w(x) bits)
    * - x.rotateRight(y)
-     - Logical right rotation, y : UInt/Int
+     - | Logical right rotation, y : UInt/Int
+       | The width of y is constrained to the width of log2Up(x) or less
      - T(w(x) bits)
    * - x.clearAll[()]
      - Clear all bits
@@ -150,15 +152,12 @@ Logic
 
 .. note::
 
-   ``x rotateLeft y`` and ``x rotateRight y`` are also valid syntax.
-
-.. note::
-
    Notice the difference in behaviour between ``x >> 2``:T(w(x)-2) and ``x >> U(2)``:T(w(x))
-   due to the Scala type of ``y``.
+   due to the Scala type of :code:`y`.
 
    The difference is that in the first case 2 is an ``Int`` (which can be seen as an
-   "elaboration integer"), and in the second case it is a hardware signal (type ``UInt``).
+   "elaboration integer constant"), and in the second case it is a hardware signal
+   (type ``UInt``) that may or may-not be a constant.
 
 .. code-block:: scala
 


andreasWallner:
If you have a request for improvement please don't create a PR with a "please fix me" - create an Issue instead.

dlmiles:
Sorry for the place-holder.  Looks like I had already added a NOTES section drawing attention to the `y` parameter type.

After another reading of the area these changes should further improve clarity.

undrafted

source/SpinalHDL/Data types/Int.rst
```diff
@@ -150,15 +150,22 @@ Logic
...
    Notice the difference in behaviour between ``x >> 2``:T(w(x)-2) and ``x >> U(2)``:T(w(x))
    due to the Scala type of ``y``.
 
    The difference is that in the first case 2 is an ``Int`` (which can be seen as an
-   "elaboration integer"), and in the second case it is a hardware signal (type ``UInt``).
+   "elaboration integer constant"), and in the second case it is a hardware signal
+   (type ``UInt``) that may or may-not be a constant.  In this example it is a
+   constant described as ``U(2)`` but if the value was provided from a
+   ``Reg(UInt(2 bits))`` or other similar dynamic signal that would elborate
+   differently and has appropiate width constraints upon `y` in sympathy with the
```
andreasWallner, line 167: There are no constraints placed on y, see:
```scala
object Blub extends App {
  SpinalVerilog(new Component {
    val ui = in port UInt(5 bit)
    val shift = in port UInt(20 bit)
    val uol = ui >> U(2)
    val uos = ui >> shift

    uol.asOutput()
    uos.asOutput()
  })
}
```
Which builds perfectly fine (even though shift is 20 bit).

Also: I would argue that both cases elaborate the same, a signal is being used for shifting, there will be a difference after synthesis.

Nits:
- In this example it is a constant described as ``U(2)`` -> simplify to `While U(2) is a constant, if...`
- `y` and `x` have wrong code tags?
- `in sympathy with the target x being modified` -> `depending on x`?

andreasWallner (changes requested):

source/SpinalHDL/Data types/Int.rst
dlmiles: Just to offer the example where I did width constraint placed upon the part `y` by SpinalHDL for the rotates:

```
SpinalVerilog(new Component {
    val ui = in port Bits(4 bit)
    val shift = in port UInt(4 bit)  // This needs to be log2Up(ui.getWidth) or less
    val someBits = Bits()
    someBits := ui.rorateLeft(shift)
    // someBits := ui << shift
  })
```

Will re-read and work out what needs clarity shortly.

source/SpinalHDL/Data types/Int.rst
```diff
@@ -150,15 +150,22 @@ Logic
 
 .. note::
 
-   ``x rotateLeft y`` and ``x rotateRight y`` are also valid syntax.
+   ``x rotateLeft y`` and ``x rotateRight y`` are also valid syntax.  Due to
+   the Scala languge able to infer a method with argument invocation from
```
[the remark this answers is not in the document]
dlmiles: Do we know what is the name Scala ascribes to the mechanism in those specific cases ?  inferred maybe the wrong label here, does not exactly feel declarative if some situations exists where it does not occur, but with the addition of parenthesis and a full-stop character it would always occur.

Maybe `interpret the meaning of`  would replace the word `infer`.

The purpose of the statement is to inform a new user to Scala language that the cause of the difference in notation is an aspect of the Scala language not SpinalHDL and there is an equivalence of purpose.  So the reader knows at that point they can research elsewhere (outside of SpinalHDL docs) for more info about that.

source/SpinalHDL/Data types/Int.rst
dlmiles: Moved width constraint being explained to more specific location near respective API

source/SpinalHDL/Data types/Int.rst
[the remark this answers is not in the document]
andreasWallner: Scala calls it "method invocation" https://docs.scala-lang.org/style/method-invocation.html

source/SpinalHDL/Data types/Int.rst
andreasWallner: So the only remaining point is that both versions still elaborate the same

source/SpinalHDL/Data types/Int.rst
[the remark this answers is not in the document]
dlmiles: +1 to remove it then

source/SpinalHDL/Data types/Int.rst
dlmiles: This also related to the rotate methods, but the change in elaboration isn't anything that optimization from pruning wouldn't achieve so I removed that also.

andreasWallner (approved):

andreasWallner:
Thanks!

PR #190, opened 2023-06-04
Data types/bundle: updated (generate keyword)

dlmiles:

https://dlmiles.github.io/SpinalDoc-RTD/docs_multi/html/vTEST/SpinalHDL/Data%20types/bundle.html

See review marker looking for assistance to reword this area. Thanks

### DIFF
--- a/source/SpinalHDL/Data types/bundle.rst
+++ b/source/SpinalHDL/Data types/bundle.rst
@@ -45,6 +45,10 @@ Unless ``dataWidth`` is greater than 0, there will be no ``data`` signal in elab
      val data = (dataWidth > 0) generate (UInt(dataWidth bits))
    }
 
+.. note::
+
+   See also :ref:`generate <generate>` for information about this SpinalHDL method.
+
 Operators
 ^^^^^^^^^
 
--- a/source/SpinalHDL/Structuring/parametrization.rst
+++ b/source/SpinalHDL/Structuring/parametrization.rst
@@ -4,47 +4,123 @@
 Parametrization
 ==================
 
-Introduction
-------------
+There are multiple aspects to parametrization :
 
-There is multiple aspect to parametrization : 
+- Providing and the management of, elaboration time parameters provided
+  to SpinalHDL during elaboration of the design
+- Using the parameter data to allow the designer to perform any kind
+  of hardware construction, configuration and interconnection task
+  needed in the design.  Such as optional component generation within
+  the hardware design.
 
-- Providing elaboration time parameters to the design
-- Optionaly generate some hardware
+Parallels exist with the aims of HDL features such as Verilog module
+parameters and VHDL generics.  SpinalHDL brings a far richer and more
+powerful set of capabilities into this area with the additional
+protection of Scala type safety and SpinalHDL built in HDL design rule
+checking.
+
+The SpinalHDL mechanisms for parameterization of components is not built
+on top of any native HDL mechanism and so is not impeded by HDL language
+level/version support or restrictions about what can be achieved in hand
+written HDL.
+
+For readers looking to interoperate with parameterized Verilog or
+genericized VHDL using SpinalHDL, please see the section on :ref:`BlackBox <BlackBox>`
+IP for those scenarios your project requires.
+
+
 
 Elaboration time parameters
 ------------------------------------------
 
-You can use the whole scala syntax to provide elaboration time parameters. 
+You can use the whole Scala syntax to provide elaboration time parameters.
+
+The whole syntax means you have the entire power and feature set of the
+Scala language at your disposal to solve parameterization requirements for
+your project at the level of complexity you choose.
+
+SpinalHDL does not place any opinionated restrictions on how to achieve
+your parameterization goals.  As such there are many Scala design patterns
+and a few SpinalHDL helpers that can be used to manage parameters that
+are suited to different parameter management scenarios.
+
+Here are some examples and ideas of the possibilities:
+
+ * Hardwired code and constants (not strictly parameter management at all
+   but serves to hilight the most basic mechanism, a code change, not a
+   parameter data change)
+ * Constant values provided from a companion object that are static
+   constants in Scala.
+ * Values provided to Scala class constructor, often a ``case class`` that
+   causes Scala to capture those constructor argument values as constants.
+ * Regular Scala flow-control syntax, not limited to but including
+   conditionals, looping, lambdas/monads, everything.
+ * Config class pattern (examples exist in library items such as
+   UartCtrlConfig_, SpiMasterCtrlConfig)
+ * Project defined 'Plugin' pattern (examples exist in the VexRiscV_ project
+   to configure the feature set the resulting CPU IP core is built with)
+ * Values and information loaded from a file or network based source, using
+   standard Scala/JVM libraries and APIs.
+ * `any mechanism you can create`
+
+All of the mechanisms result in a change in resulting elaborated HDL output.
+
+This could vary from a single constant value change all the way through to
+describing the entire bus and interconnection architecture of an entire SoC
+all without leaving the Scala programming paradigm.
+
 
 Here is an example of class parameters
 
 .. code-block:: scala
 
-  case class MyBus(width : Int) extends Bundle{
+  case class MyBus(width : Int) extends Bundle {
     val mySignal = UInt(width bits)
   }  
   
 .. code-block:: scala
 
-  case class MyComponent(width : Int) extends Component{
+  case class MyComponent(width : Int) extends Component {
     val bus = MyBus(width)
   }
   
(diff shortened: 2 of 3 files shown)


dlmiles:
rebased onto current master, this commit contains: https://github.com/SpinalHDL/SpinalDoc-RTD/compare/master..0af4814447ac2ab558c749a86c1c51f63f6d0e80

undrafted

andreasWallner:
Before I take a detailed look, have you seen https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Structuring/parametrization.html#optional-hardware ?
That explains `generate` with some examples - if we feel the need to explain generate we should link there.

dlmiles:
The parametrization.html is a better place to put the details for `generate` with a link from the location in the commit.

While I have read parametrization.html to check any obvious proof-reading matters, I didn't recall that is was covering `generate`.

I did not find the `Structuring/parametrization.html#optional-hardware` page as inspirational as I would expect, considering IMHO one of the major benefits of SpinalHDL is parameter management.  It is heavy on examples and light on exposing an entire new genre to the world of HDL that SpinalHDL can provide with ease.

reworking both areas to see what can be done...

dlmiles:
Initial draft to attempt to add some background and inspiration to what the content conveys, then lead into examples.

source/SpinalHDL/Data types/bundle.rst
```diff
@@ -45,6 +45,10 @@ Unless ``dataWidth`` is greater than 0, there will be no ``data`` signal in elab
      val data = (dataWidth > 0) generate (UInt(dataWidth bits))
    }
 
+.. note::
+
+   See also :ref:`generate <generate>` for information about this SpinalHDL keyword like directive.
```
andreasWallner, line 50:
```suggestion
   See also :ref:`generate <generate>` for information about this SpinalHDL method.
```

source/SpinalHDL/Structuring/parametrization.rst
```diff
@@ -15,7 +15,41 @@ There is multiple aspect to parametrization :
 Elaboration time parameters
 ------------------------------------------
 
-You can use the whole scala syntax to provide elaboration time parameters. 
+You can use the whole Scala syntax to provide elaboration time parameters.
+
+The whole syntax means you have the entire power and feature set of the
+Scala language at your disposal to solve parametrization requirements for
```
andreasWallner, line 21:
```suggestion
Scala language at your disposal to solve parameterization requirements for
```

source/SpinalHDL/Structuring/parametrization.rst
```diff
@@ -15,7 +15,41 @@ There is multiple aspect to parametrization :
...
+   but serves to hilight the most basic mechanism, a code change, not
+   paramater data change)
+ * Constant values provided from a companion object that are static
+   constants in Scala.
+ * Values provided to Scala class constructor, often a ``case class`` that
+   causes Scala to capture those constructor argument values as constants.
+ * Regular Scala flow-control syntax, not limited to but including
+   conditionals, looping, everything.
+ * Config class pattern (examples exist in library items such as
+   UartCtrlConfig, SpiMasterCtrlConfig)
```
andreasWallner, line 41: we could link this to e.g. https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Examples/Intermediates%20ones/uart.html#controller-construction-parameters

source/SpinalHDL/Structuring/parametrization.rst
```diff
@@ -15,7 +15,41 @@ There is multiple aspect to parametrization :
...
+ * Plugins (examples exist in VexRiscV project to configure the feature set
+   the resulting CPU IP core is built with)
+ * Values and information looked from a file or remote based source.
```
andreasWallner, line 44:
```suggestion
 * Values and information read from a file or remote based source.
```

source/SpinalHDL/Structuring/parametrization.rst
```diff
@@ -31,13 +65,17 @@ Here is an example of class parameters
     val bus = MyBus(width)
   }
   
-You can also use global variable defined in scala object's, but note that recently was added the ScopeProperty feature which improve on that solution.
+You can also use global variable defined in Scala object's (companion object
```
andreasWallner, line 68: I have commented on that in #193 - let's discuss that there?

source/SpinalHDL/Structuring/parametrization.rst
```diff
@@ -46,6 +84,21 @@ For optional signal :
     val mySignal = flag generate (Bool())  //Equivalent to "if (flag) in Bool() else null"
   }
 
+.. note::
```
andreasWallner, line 87: Why not make this normal text?

source/SpinalHDL/Structuring/parametrization.rst
```diff
@@ -46,6 +84,21 @@ For optional signal :
...
+
+   The ``generate`` keyword like directive is a mechanism to evaluate the expression
+   that follows for a value (in this case the construction of a Bool() type), if the
+   previous predicate expression evaluated true (in this case when `flag` is
+   true).  If the predicate expression evalutes to false then a null
+   value is the result and would be assigned to the Scala val `mySignal`.
+   This maybe used in cases to help parameterize the SpinalHDL hardware
```
andreasWallner, line 94:
```suggestion
   This may be used in cases to help parameterize the SpinalHDL hardware
```

source/SpinalHDL/Structuring/parametrization.rst
```diff
@@ -46,6 +84,21 @@ For optional signal :
     val mySignal = flag generate (Bool())  //Equivalent to "if (flag) in Bool() else null"
   }
 
+.. note::
+
```
andreasWallner, line 88: - `generate` is a method/function, not a "keyword like directive" - `flag.generate(Bool())` is exactly the same
- "follows for a value" seemed like a typo to me (reading it a few times I get what you want to say) + this is written quite complicated. How about something along the lines of "generate will evaluate the given expression depending on the predicate, returning `null` if the predicate is false"?

andreasWallner (changes requested):

source/SpinalHDL/Structuring/parametrization.rst
dlmiles: Technically everything is a method/function, no code exists in the JVM that doesn't follow that paradigm of a bytecode entry point.
So the question then is, what would you want to initially teach someone who does not know anything about the subject ?
Can this be done in a way that they grasp the new concept enough to understand it and move to the next thing.
I find the term `keyword like directive` to encapsulate the nature of this one, it looks and acts like the keyword, it is controlling behaviour like a directive, the syntax is expected to be used like a keyword.   The term stays in the domain of describing how to use it, and not what it does.

Agree on the "follows for a value" is not a good explanation, I am trying to emphasis the expression must return a value, not `Unit` / void.

With your wording, it sounds funny to start the explanation about the thing that happens 2nd (evaluation of expression when predicate true).  So starting the explanation with the evaluation of the predicate should be first, which also follows how it is evaluated if there are additional side-effects (other than the value returned) during the optionally evaluated expression when predicate true.  Changing the order/rewording would solve that.

Maybe it can be concisely explained in terms of a `Ternary` operator concepts ? 

The first sentence (or so) is trying to formally define what is does in terms and rules that are easy to work through and difficult to misunderstand.  I don't know if that adds to the perceived complexity of the explanation.

andreasWallner: You might find "keyword like directive" to be a good term, but this is not used anywhere else in either Scala or SpinalHDL docs so googling for it you find no helpful search results. Inventing new names will not make this any easier - and might obfuscate what is actually happening.

Back to the wording, let's turn it around then: "If the predicate is true, generate will evaluate the given expression and return the result, otherwise is returns null"?

source/SpinalHDL/Structuring/parametrization.rst
dlmiles: Went with `loaded from a file or network based source`

source/SpinalHDL/Structuring/parametrization.rst
dlmiles: I like the term it conveys so much to me.

Give this version a go.

I think everything in Scala returns a value, it is the Monadic way, even Unit and null is a value.
But the purpose of referencing it is to make the user think about the value being returned, but maybe that is a given from the code example and expected Scala understanding of the reader.

andreasWallner (changes requested):
LGTM Just one small typo (I think) and did you see that the BlackBox reference generates a warning?

dlmiles:
"BlackBox reference generates a warning"

Can check for any missing refs after all commits merged.  The BlackBox ref should be in Tranche 4.

andreasWallner (approved):

andreasWallner (approved):

andreasWallner:
Thanks!

PR #202, opened 2023-07-08
Add StreamTransactionExtender explanation.

Readon:


source/SpinalHDL/Libraries/stream.rst
```diff
@@ -534,6 +534,27 @@ This util take its input stream and routes it to ``outputCount`` stream in a seq
...
+.. code-block:: scala
+
+   val inputStream = Stream(Bits(8 bits))
+   val outputStream = Stream(Bits(8 bits))
+   val count = UInt(3 bits)
+   val extender = StreamTransactionExtender(inputStream, outputStream, count) {
+       (id, payload, last) => payload
+   }
+
+This ``extender`` have provided several status signals, such as ``working``, ``last``, ``done`` where ``working`` means there is one transaction accepted and on going, ``last`` stands for waiting the last output stream fire, ``done`` represent the last output stream fire and the current input transcation process is done.
```
dlmiles, line 552: This ``extender`` provides several status signals, such as ``working``, ``last``, ``done`` where ``working`` means there is one input transaction accepted and in-progress, ``last`` indicates when the last output stream has fired (started), ``done`` represents the last output stream has fired and become valid making the current input transaction process is complete and ready to start another transaction.

Note: the ``last`` is not clear if it indicates the last output stream has been seen/queued/requested, or has started, or has completed (I don't think it has this meaning).   Would it be true to say that ``last`` fires one cycle before ``done`` is expected ?

source/SpinalHDL/Libraries/stream.rst
```diff
@@ -534,6 +534,27 @@ This util take its input stream and routes it to ``outputCount`` stream in a seq
      outputCount = 3
    )
 
+StreamTransactionExtender
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This util helps to extend one input transaction with specified ``count``+1 times of output transactions with the same payload.
```
dlmiles: This utility will take one input transaction and generate one output transaction, it provides the facility to repeat the payload value ``count``+1 times repeated into the same output transaction.

The summary doesn't make it immediately understood what is does.
Use of the term 'helps' makes it sound like it is a partial solution to a problem and is not the free standing solution that it is.

Can we provide a wavedrom of an example of the signalling ?

source/SpinalHDL/Libraries/stream.rst
```diff
@@ -534,6 +534,27 @@ This util take its input stream and routes it to ``outputCount`` stream in a seq
...
+The ``count`` signal would be captured and registered while inputStream fire.
+
```
dlmiles, line 542: A ``count`` value of zero would result in no additional payload insertion causing the input and output transactions to be identical.

source/SpinalHDL/Libraries/stream.rst
```diff
@@ -534,6 +534,27 @@ This util take its input stream and routes it to ``outputCount`` stream in a seq
...
+.. code-block:: scala
+
+   val inputStream = Stream(Bits(8 bits))
+   val outputStream = Stream(Bits(8 bits))
+   val count = UInt(3 bits)
+   val extender = StreamTransactionExtender(inputStream, outputStream, count) {
+       (id, payload, last) => payload
```
dlmiles, line 549: It is not immediately obvious what this line is doing, it also introduces 'id'.  Is that an ordinal index of the count ?  is it 0-based or 1-based, is it 0 to count+1 ?  is `last` from the inputStream's view ?

Maybe a comment will help, or other potential code examples as commented out code will clarify what is possible/allowed/disallowed here.

source/SpinalHDL/Libraries/stream.rst
dlmiles: Question: the `count` that is captured, is it captured once, or at every inputStream payload valid signal ?  That should be stated.  Maybe like:

The ``count`` is captured and registered each time inputStream fires for an individual payload.

source/SpinalHDL/Libraries/stream.rst
Readon: Normally we have 0-based index. ``id`` is just the count value, shows the index of output transfers.

source/SpinalHDL/Libraries/stream.rst
```diff
@@ -534,6 +534,40 @@ This util take its input stream and routes it to ``outputCount`` stream in a seq
      outputCount = 3
    )
 
+StreamTransactionExtender
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This utility will take one input transfer and generate serval output transfers, it provides the facility to repeat the payload value ``count+1`` times into output transfers.
```
dlmiles, line 540: several

source/SpinalHDL/Libraries/stream.rst
dlmiles: Maybe just a comment around it then:

```
   // id, is the 0-based index of total output transfers so far from this input transaction
   // last, is the current inputStream.last indication
   // the returned payload is allowed to be modified
   (id, payload, last) => payload
```

In this context it would be really nice to expose the `repeatIndex` with 0 being the original and 1 being the first repeat, etc...  with it resetting back to zero for the next payload eaten from inputStream, I am assuming it lets you modify the payload in there.

But maybe doing that has a hardware cost when the feature is not used ?  Which would make it a bad idea.

dlmiles:
LGTM

PR #203, opened 2023-07-09
Tranche 4

dlmiles:


source/SpinalHDL/Other language features/utils.rst
```diff
@@ -59,6 +59,7 @@ You can get more information about how hardware data types are managed on the :r
 
 .. note::
    If you use the ``cloneOf`` function on a ``Bundle``, this ``Bundle`` should be a ``case class`` or should override the clone function internally.
+   REVIEWME You can't say that and not provide an example clone() implementation / details of considerations with the implementation.
```
dlmiles, line 62: Note to self, still need to provide this.

source/SpinalHDL/Libraries/fiber.rst
```diff
@@ -10,22 +10,23 @@ Currently in developpement.
...
+- Automatic requirement negotiation between multiple agents in a decentralized way, ex : between masters and slaves of a memory bus
 
 The framework is mainly composed of : 
 
 - Handle[T], which can be used later to store a value of type T. 
 - handle.load which allow to set the value of a handle (will reschedule all tasks waiting on it)
 - handle.get, which return the value of the given handle. Will block the task execution if that handle isn't loaded yet
 - Handle{ code }, which fork a new task which will execute the given code. The result of that code will be loaded into the Handle
-- soon(handle), which allow the current task to announce that soon it will load that handle with a value (used to track which handle will
+- soon(handle), which allow the current task to announce that soon it will load that handle with a value (used to track which handle will ???REVIEWME)
```
dlmiles, line 26: Note to self, still need to provide this.

source/SpinalHDL/Other language features/utils.rst
andreasWallner: I think we can: The preferred way should be to make it a `case class`. AFAIR we haven't needed to make a Bundle a `class` for `lib` - I'd not spend much space here to describe something that most users will never need a can be done automatically.
I thought about linking to the official docs here, but they are not much more helpful: https://www.scala-lang.org/api/current/scala/collection/mutable/Cloneable.html
But the link might still be better then just mentioning `clone` without the trait or anything

source/SpinalHDL/Simulation/install/Verilator.rst
```diff
@@ -5,6 +5,14 @@ Setup and installation of Verilator
 
 SpinalSim + Verilator is supported on both Linux and Windows platforms.
 
+It is recommended that v4.218 is the oldest Verilator version to use.  While it maybe
+possible to use older verilator versions, some optional and Scala source dependant
```
andreasWallner, line 9:
```suggestion
possible to use older verilator versions, some optional and Scala source dependent
```

source/SpinalHDL/Simulation/install/Verilator.rst
```diff
@@ -5,6 +5,14 @@ Setup and installation of Verilator
...
+features that SpinalHDL can use (such as Verilog $urandom support) may not be supported
```
andreasWallner, line 10:
```suggestion
features that SpinalHDL can use (such as Verilog ``$urandom`` support) may not be supported
```

andreasWallner (changes requested):
LGTM in almost all places, just some minor points

source/SpinalHDL/Libraries/fiber.rst
dlmiles: I do not know the fiber system well enough to comment.

This line of the original description did not help to understand the purpose of soon(handle).

Your new description explains the purpose of part of the API, so this is better :)

source/SpinalHDL/Other language features/utils.rst
dlmiles: Can we emit a code fragment example with this ?

In particular the `should override the clone function internally` is the part that needs to be clarified.
The example would confirm the exact interface/trait being provided and what a valid looking implementation can look like, even if this seems obvious to you, seeing it in code helps clarify to the reader it is nothing special.

```
class MyBundle() extend Bundle {
  // when not a case class there is no implicit clone() implementation ???
  val signalOne = Bool()
  val signalTwo = UInt(4 bits)
  override def clone(): this.type = {
    new MyBundle()  // ????
  }
}

val inst = cloneOf(bund)

```

Dolu1990: Here is an example : 

```scala
    class MyBundle(ppp : Int) extends Bundle {
      val a = UInt(ppp bits)
      override def clone = new MyBundle(ppp)
    }
    val x = new MyBundle(3)
    val typeDef = HardType(new MyBundle(3))
    val y = typeDef()

    cloneOf(x) //need clone methode, else it crash
    cloneOf(y) //Is ok
```

Note that HardType fix the issue, so, it also fix it when used for Stream / Vec / ... type parameters

dlmiles: This is on preview here: https://dlmiles.github.io/SpinalDoc-RTD/docs_multi/html/latest/SpinalHDL/Other%20language%20features/utils.html#cloning-hardware-datatypes

andreasWallner (approved):

andreasWallner:
looks good to me, can we merge this? (I don't want to merge mine first and force you to do the rebase/merge ;-) )

source/SpinalHDL/Other language features/utils.rst
Dolu1990: > If you use the cloneOf function on a Bundle, this Bundle should be a case class or should override the clone function internally.

Hmm that's not realy true, as long as your class has no parameter, that's fine :)

Dolu1990: Maybe it could be : 
"The cloneOf(Bundle) function can only be used on Bundle which either have no parameters, have the clone function defined or are case class."

Dolu1990:
<3

PR #204, opened 2023-07-24
add slots demo

Dolu1990:
This demo show an alternative to Vec when it is about defining repetitive hardware structure.

### DIFF
--- a/examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
+++ b/examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
@@ -1,0 +1,68 @@
+package spinaldoc.examples.advanced
+
+import spinal.core._
+import spinal.lib._
+import scala.language.postfixOps
+
+case class SlotsDemo(slotsCount : Int) extends Component {
+  //...
+  
+  
+  //Create the hardware for each slot
+  //Note each slot is an Area, not a Bundle
+  val slots = for(i <- 0 until slotsCount) yield new Area{
+    //Because the slot is an Area, we can define mix signal, registers, logic definitions
+    //Here are the registers for each slots
+    val valid = RegInit(False)
+    val address = Reg(UInt(8 bits))
+    val age = Reg(UInt(16 bits)) //Will count since how many cycles the slot is valid
+
+    //Here is some hardware behaviour for each slots
+    //Implement the age logic
+    when(valid){
+      age := age + 1
+    }
+
+    //removeIt will be used as a slot interface later on
+    val removeIt = False
+    when(removeIt){
+      valid := False
+    }
+  }
+
+  //Logic to allocate a new slot
+  val insert = new Area{
+    val cmd = Stream(UInt(8 bits)) //interface to issue requests
+    val free = slots.map(!_.valid)
+    val freeOh = OHMasking.first(free) //Get the first free slot (on hot mask)
+    cmd.ready := free.orR //Only allow cmd when there is a free slot
+    when(cmd.fire){
+      //slots.onMask(freeOh)(code) will execute the code for each slot where the corresponding freeOh bit is set
+      slots.onMask(freeOh){slot =>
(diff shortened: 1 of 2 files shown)


examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
```diff
@@ -0,0 +1,68 @@
...
+  val slots = for(i <- 0 until slotsCount) yield new Area{
+    //Because the slot is an Area, we can define mix signal, registers, logic definitions
+    //Here are the registers for each slots
+    val valid = RegInit(False)
+    val address = Reg(UInt(8 bits))
+    val age = Reg(UInt(16 bits)) //Will count since how many cycles the slot is valid
+
+    //Here is some hardware behaviour for each slots
+    //Implement the age logic
+    when(valid){
```
andreasWallner, line 22: It doesn't change much about the example but should there be a check against overflow? (possibly using a Counter for the age)?

examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  //Logic to allocate a new slot
+  val insert = new Area{
+    val cmd = Stream(UInt(8 bits)) //interface to issue requests
+    val free = slots.map(!_.valid)
+    val freeOh = OHMasking.first(free) //Get the first free slot (on hot mask)
+    cmd.ready := free.orR //Only allow cmd when there is a free slot
+    when(cmd.fire){
+      //slots.onMask(freeOh)(code) will execute the code for each slot where the corresponding freeOh bit is set
+      slots.onMask(freeOh){slot =>
```
andreasWallner, line 41: I think I'll add `onMask` to e.g. the tools, that warrants a separate explanation...

examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
```diff
@@ -0,0 +1,68 @@
...
+    val reader = slots.reader(oh) //Create a facility to read the slots using "oh" as index
+    val age = reader(_.age) //Age of the slot which is selected by "oh"
+  }
+  
+  //...
+}
+
+object SlotsDemo extends App {
+  SpinalVerilog(SlotsDemo(4))
+}
```
andreasWallner, line 68: It would be cool to also have a usage example here

examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
andreasWallner: I thought about using `+|` mostly about avoiding confusion why we are doing it, but I think it's fine either way.

Dolu1990: Ahhh i forgot about the `+|` XDXD
hmmm still, just better focussing on the example core than adding too much distractions :D

examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
Dolu1990: Yes `onMask` and `onSel` deserve proper doc XD

examples/src/main/scala/spinaldoc/examples/advanced/Slots.scala
Dolu1990: So, for instance, i'm using this slot stuff for the Tilelink coherency hub : 
https://github.com/SpinalHDL/SpinalHDL/blob/008c73f1ce18e294f137efe7a1442bd3f8fa2ee0/lib/src/main/scala/spinal/lib/bus/tilelink/coherent/Hub.scala#L376

It keep track of all ongoing memory probes in flight.

andreasWallner (approved):

mrcmry:
Deleted merged branch for cleanup.

PR #208, opened 2023-08-22
Clarify subdivide

andreasWallner:
This started out just wanting to move subdivide to the access functions, but now ended in a few additional changes. Among the bigger ones that e.g. `abs` and similar functions are now listed as "conversion" functions, some additional examples and a fix to the output order of `subdivideIn` in the examples.

### DIFF
--- a/source/SpinalHDL/Data types/Int.rst
+++ b/source/SpinalHDL/Data types/Int.rst
@@ -6,7 +6,8 @@ UInt/SInt
 Description
 -----------
 
-The ``UInt``/``SInt`` type corresponds to a vector of bits that can be used for signed/unsigned integer arithmetic.
+The ``UInt``/``SInt`` types are a vectors of bits interpreted as two's complement unsigned/signed integers.
+They can do what ``Bits`` can do, with the addition of unsigned/signed integer arithmetic and comparisons.
 
 Declaration
 -----------
@@ -15,43 +16,33 @@ The syntax to declare an integer is as follows:  (everything between [] is optio
 
 .. list-table::
    :header-rows: 1
-   :widths: 5 10 2
+   :widths: 5 10
 
    * - Syntax
      - Description
-     - Return
    * - | UInt[()]
        | SInt[()]
      - Create an unsigned/signed integer, bits count is inferred
-     - | UInt
-       | SInt
    * - | UInt(x bits)
        | SInt(x bits)
      - Create an unsigned/signed integer with x bits
-     - | UInt
-       | SInt
    * - | U(value: Int[,x bits])
        | U(value: BigInt[,x bits])
        | S(value: Int[,x bits])
        | S(value: BigInt[,x bits])
      - Create an unsigned/signed integer assigned with 'value'
-     - | UInt
-       | SInt
    * - | U"[[size']base]value"
        | S"[[size']base]value"
-     - Create an unsigned/signed integer assigned with 'value' (Base : 'h', 'd', 'o', 'b')
-     - | UInt
-       | SInt
-   * - | U([x bits,] :ref:`element <element>`, ...)
-       | S([x bits,] :ref:`element <element>`, ...)
-     - Create an unsigned integer assigned with the value specified by elements
-     - | UInt
-       | SInt
+     - | Create an unsigned/signed integer assigned with 'value'
+       | (base: 'h', 'd', 'o', 'b')
+   * - | U([x bits,] elements: Element*)
+       | S([x bits,] elements: Element*)
+     - Create an unsigned integer assigned with the value specified by :ref:`elements <element>`
 
 .. code-block:: scala
 
-   val myUInt = UInt(8 bits)
-   myUInt := U(2,8 bits)
+   val myUInt = UInt(8 bit)
+   myUInt := U(2, 8 bit)
    myUInt := U(2)
    myUInt := U"0000_0101"  // Base per default is binary => 5
    myUInt := U"h1A"        // Base could be x (base 16)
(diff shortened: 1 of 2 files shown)


source/SpinalHDL/Data types/Int.rst
```diff
@@ -4,54 +4,45 @@ UInt/SInt
 =========
 
 Description
-^^^^^^^^^^^
+-----------
 
-The ``UInt``/``SInt`` type corresponds to a vector of bits that can be used for signed/unsigned integer arithmetic.
+The ``UInt``/``SInt`` types are a vectors of bits interpreted as two's complement unsigned/signed integers.
```
dlmiles, line 9: `are a vectors of bits` maybe `are a vector of bits` or `are vectors of bits` (preferred).

source/SpinalHDL/Data types/Int.rst
```diff
@@ -338,52 +345,136 @@ To cast a ``Bool``, a ``Bits``, or an ``SInt`` into a ``UInt``, you can use ``U(
...
+     - Access a :ref:`range <range>` of bits. Ex : myBits(4 downto 2)
+     - Bits(range.size bits)
+   * - x.subdivideIn(y slices, [strict: Boolean])
+     - Subdivide x into y slices, y: Int
+     - Vec(Bits(...), y)
+   * - x.subdivideIn(y bits, [strict: Boolean])
+     - Subdivide x in multiple slices of y bits, y: Int
+     - Vec(Bits(y bit), ...)
+   * - x.msb
+     - Access most significant bit of x (highest index)
```
dlmiles, line 405: Is it worth pointing out how this might vary or not-vary depending upon if x is signed (SInt) or unsigned (UInt/Bits) ?

That is not obvious from reading the docs, especially when `mySInt.sign` is just below.

source/SpinalHDL/Data types/Int.rst
```diff
@@ -338,52 +345,136 @@ To cast a ``Bool``, a ``Bits``, or an ``SInt`` into a ``UInt``, you can use ``U(
...
+
+   // read / assign the msb / leftmost bit / x.high bit
+   val isNegative = mySInt_16bit.sign
+   myUInt_16bit.msb := False
+
+Subdivide details
+"""""""""""""""""
+
+Both overloads of ``subdivideIn`` have an optional parameter ``strict`` (i.e. ``subdivideIn(slices: SlicesCount, strict: Boolean = true)``).
+If ``strict`` is ``true`` an error will be raised if the input could not be divided evenly. If set to ``false`` the generated pieces may
```
dlmiles, line 448: `evenly` maybe `exactly` ?   use of even and odd terms start to become ambiguous when the slices!=2

`... the generated pieces may vary in size as necessary.`
`... the generated pieces may vary the size of the last slice as necessary.`

source/SpinalHDL/Data types/bits.rst
```diff
@@ -242,52 +244,112 @@ To cast a ``Bool``, ``UInt`` or an ``SInt`` into a ``Bits``, you can use ``B(som
...
+
+   // read / assign the msb / leftmost bit / x.high bit
+   val isNegative = myBits_16bit.msb
+   myBits_16bit.msb := False
+
+Subdivide details
+"""""""""""""""""
+
+Both overloads of ``subdivideIn`` have an optional parameter ``strict`` (i.e. ``subdivideIn(slices: SlicesCount, strict: Boolean = true)``).
+If ``strict`` is ``true`` an error will be raised if the input could not be divided evenly. If set to ``false`` the generated pieces may
```
dlmiles, line 323: DUPLICATE COMMENT AS PER Int.rst FILE:444

`evenly` maybe `exactly` ?   use of even and odd terms start to become ambiguous when the slices!=2

`... the generated pieces may vary in size as necessary.`
`... the generated pieces may vary the size of the last slice as necessary.`

source/SpinalHDL/Data types/Int.rst
Dolu1990: "not be divided evenly" could be instead "not be divided in equal parts"

"the generated pieces may have varying size if necessary" hmm the intended use is that the last element will be the one which "adapts"
So 
Could be : "the last elements will have the remainder"

andreasWallner: done and changed the last sentence to
> If set to ``false`` the last element may be smaller than the other (equal sized) elements.

source/SpinalHDL/Data types/bits.rst
andreasWallner: as above

source/SpinalHDL/Data types/Int.rst
```diff
@@ -338,52 +345,136 @@ To cast a ``Bool``, a ``Bits``, or an ``SInt`` into a ``UInt``, you can use ``U(
...
+    val pieces = output8.subdivideIn(2 slices)
+    // assign to output8
+    pieces(0) := 0xf
+    pieces(1) := 0x5
 
 Misc
-~~~~
+^^^^
+
+The operations listed below that create hardware signals all create new signals.
```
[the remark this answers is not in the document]
andreasWallner: Removed the whole sentence - didn't really like it and I think the following one should be clear enough.

source/SpinalHDL/Data types/Int.rst
dlmiles: Is that identical to `x.sign` in the case of SInt types ?  or is that the highest MSB of the range `x(x.getWidth - 1)` ?
That is the clarification to make.

andreasWallner: added that as a note

Dolu1990:
Thanks ^^

PR #209, opened 2023-08-22
Fix VCS blackbox simulation notes

andreasWallner:


dlmiles:
LGTM correction for https://github.com/dlmiles/SpinalHDL/blob/dev/core/src/main/scala/spinal/core/Spinal.scala#L325

PR #210, opened 2023-09-14
Update Fix.rst comparison table

callistopili:
Updated a fix for the table rappresenting the Logic operation with Fixed point. the table shows x>y twice.


andreasWallner:
Thanks for catching this!

PR #217, opened 2023-10-06
Document mem sim api

andreasWallner:
Shifted down the memory access documentation since the example uses `simPublic`.
Remove some legacy documentation and adapt headline to make memory docs easier to find (if user searches for "Memory")


Dolu1990:
<3

PR #224, opened 2023-10-20
No more convet-wrapper ?

Dolu1990:
I don't know, this line just make my setup not work, and commenting it out doesn't seems to break anything

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -295,7 +295,7 @@ if have_image_converter:
     image_converter='convert-wrapper'
 else:
     print("ERROR: convert-wrapper not found in $PATH, did you: source $PWD/bin/setup_env.sh", file=sys.stderr)
-    sys.exit(1)
+    #sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


andreasWallner:
I think the idea was to break the build in case the correct environment is not sourced (which in turn breaks the multi-version build)

Dolu1990:
thing is even when i sourced the environnement, it doesn't fix it. (on a very recent Mint distrib)

andreasWallner:
because it can't find `convert-wrapper` - weird...
could you check your PATH after loading it and whether the script is executable?

In the end I think it doesn't matter too much for a quick local run, only if one wants to build all the old versions as well... (and on CI, but there the script still seems to work as expected)

Dolu1990:
hmm i just tried again using venv, and it worked directly.
I don't know what i was doing wrong, maybe didn't sourced things properly XD

sorry for the noise XD

mrcmry:
Deleted this trivial branch not relevant anymore for cleanup.

PR #225, opened 2023-10-26
Add note that \= only works for comb signals

andreasWallner:


Dolu1990:
Thanks ^^

Dolu1990:
Note that i changed from error to warning.

PR #228, opened 2023-11-09
Add a documentation for CombInit

ronan-lashermes:
Documenting CombInit, since I stumble upon it and I did not see any description in the docs.

### DIFF
--- a/source/SpinalHDL/Semantic/assignments.rst
+++ b/source/SpinalHDL/Semantic/assignments.rst
@@ -122,7 +122,7 @@ If one is detected, it raises an error and SpinalHDL will print the path of the 
 CombInit
 --------
 
-The special ``CombInit`` method can be used to "clone" a combinatorial value for latter modification.
+``CombInit`` can be used to copy a signal and its current combinatorial assignments. The main use-case is to be able to overwrite the copied later, without impacting the original signal.
 
 .. code-block:: scala
 
--- a/source/SpinalHDL/Semantic/assignments.rst
+++ b/source/SpinalHDL/Semantic/assignments.rst
@@ -118,3 +118,51 @@ Combinatorial loops
 
 SpinalHDL checks that there are no combinatorial loops (latches) in your design.
 If one is detected, it raises an error and SpinalHDL will print the path of the loop.
+
+CombInit
+--------
+
+The special ``CombInit`` method can be used to "clone" a combinatorial value for latter modification.
+
+.. code-block:: scala
+
+    val a = UInt(8 bits)
+    a := 1
+
+    val b = a
+    when(sel) {
+        b := 2
+        //At this point, a and b are evaluated to 2 : they represent the same set of wire
+    }
+
+    val c = UInt(8 bits)
+    c := 1
+
+    val d = CombInit(c)
+    // Here c and d are evaluated to 1
+    when(sel) {
+        d := 2
+        // At this point c === 1 and d === 2.
+    }
+
+``CombInit`` clones a circuit, and initially drive it with the same input at the cloned value.
+But you can now update the circuit without impacting the initial value.
+
+If we look at the resulting Verilog, ``b`` is not present :
+
+.. code-block:: verilog
+
+    always @(*) begin
+      a = 8'h01;
+      if(sel) begin
+        a = 8'h02;
+      end
+    end
+
+    assign c = 8'h01;
+    always @(*) begin
+      d = c;
+      if(sel) begin
+        d = 8'h02;
+      end
+    end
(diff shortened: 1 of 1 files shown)


source/SpinalHDL/Semantic/assignments.rst
```diff
@@ -118,3 +118,51 @@ Combinatorial loops
 
 SpinalHDL checks that there are no combinatorial loops (latches) in your design.
 If one is detected, it raises an error and SpinalHDL will print the path of the loop.
+
+CombInit
+--------
+
+The special ``CombInit`` method can be used to "clone" a combinatorial value for latter modification.
```
andreasWallner, line 125: I'd try to distinguish this from `cloneOf` (and a small technical change since CombInit is not really a method) - a wording could be:
```suggestion
``CombInit`` can be used to copy a signal and its current combinatorial assignments. The main use-case is to be able to overwrite the copied later, without impacting the original signal.
```

source/SpinalHDL/Semantic/assignments.rst
```diff
@@ -118,3 +118,51 @@ Combinatorial loops
...
+
+.. code-block:: scala
+
+    val a = UInt(8 bits)
+    a := 1
+
+    val b = a
+    when(sel) {
+        b := 2
+        //At this point, a and b are evaluated to 2 : they represent the same set of wire
```
andreasWallner, line 135: just to align this a bit more with the usual Scala wording (a & b are references to the same object)
```suggestion
        // At this point, a and b are evaluated to 2: they reference the same signal
```

source/SpinalHDL/Semantic/assignments.rst
```diff
@@ -118,3 +118,51 @@ Combinatorial loops
...
+    c := 1
+
+    val d = CombInit(c)
+    // Here c and d are evaluated to 1
+    when(sel) {
+        d := 2
+        // At this point c === 1 and d === 2.
+    }
+
+``CombInit`` clones a circuit, and initially drive it with the same input at the cloned value.
```
andreasWallner, line 148:
```suggestion
``CombInit`` clones a circuit, and initially drive it with the same input as the cloned value.
```

source/SpinalHDL/Semantic/assignments.rst
```diff
@@ -118,3 +118,51 @@ Combinatorial loops
...
+But you can now update the circuit without impacting the initial value.
+
```
andreasWallner, line 150: We could add an example here for one of the major use-cases which is to make helper functions that break the reference to the original signal. That would explain what to use it for, I found it's pretty rare in "normal" HDL code. A dumb example would be:
```
// note that condition is an elaboration time constant
def invertedIf(b: Bits, condition: Boolean): Bits = if(condition) { ~b } else { CombInit(b) }
```
where we shouldn't return just `b` in the `false` case as that would create weird behavior (if true the return value can't influence the original b, if false it can)
Do you think that makes sense?

source/SpinalHDL/Semantic/assignments.rst
```diff
@@ -118,3 +118,51 @@ Combinatorial loops
...
+If we look at the resulting Verilog, ``b`` is not present :
```
andreasWallner, line 151: I think this is more a statement about references than `CombInit` - but I have no big issue keeping it if you think it helps a bunch.

andreasWallner (changes requested):
Thanks a bunch for adding this, I added a few suggestions.

source/SpinalHDL/Semantic/assignments.rst
ronan-lashermes: Thanks for your suggestions.
I like your invertedIf example, I was looking for a simple realistic use case : this is it.

source/SpinalHDL/Semantic/assignments.rst
ronan-lashermes: Yeah, 148 and 149 are a bit redundant with the earlier CombInit description.
151 was there to insist on the differences in the synthesis results, I would like to keep it but add a bit more contextual description.

source/SpinalHDL/Semantic/assignments.rst
ronan-lashermes: See 6222df6 for a first try.

source/SpinalHDL/Semantic/assignments.rst
ronan-lashermes: Cf 6222df6.

andreasWallner (approved):

andreasWallner:
Thanks!

Issue #201, opened 2023-07-03
simulator_specifics.rst: XSim is missing


dlmiles:
Section for `How SpinalHDL simulates the hardware with Xilinx XSim backend` is missing.

mrcmry:
Based on the conversation of #230, we can close this issue.

PR #230, opened 2023-11-10
Describe the usage of XSim.

Readon:


source/SpinalHDL/Simulation/simulator_specifics.rst
```diff
@@ -59,6 +59,26 @@ How SpinalHDL simulates the hardware with Synopsys VCS backend
...
+2. The HDL model is loaded in the simulator.
+3. The communication between the simulation and the JVM is established through shared memory. The commands are issued to the simulator using XSI.
+
+**Advantages:**
+
+* Support Xilinx built-in primitives and cores.
+
+**Limitations:**
+
+* Xilinx XSim is a **commercial** tool installed with Vivado. It is close source and not free. You have to own the licenses to **legally** use it.
```
dlmiles, line 75: Your description reads a little shocking.  "You must purchase licenses to legally use this"

... closed source and subject to licensing terms to use.   While development with Xilinx higher performance chips may require a non-free versions of the Vivado tooling suite to target those FPGAs. Many Xilinx based bottom tier development boards (Diligent Arty / Zybo series and equivalent competing products) provide access to obtain a free license for the Vivado suite that is capable of targeting the lower end FPGAs.

source/SpinalHDL/Simulation/simulator_specifics.rst
```diff
@@ -59,6 +59,26 @@ How SpinalHDL simulates the hardware with Synopsys VCS backend
...
+* Vivado versions prior to 2019.1 do not work properly.
+
+Before using XSim as the simulation backend, make sure that you have done following steps.
+1. Define VIVADO_HOME environment variable to specify where your vivado located. ex `export VIVADO_HOME=/d/Xilinx/Vivado/2022.1` (under MSYS2).
+2. Make sure two vivado path is inside the PATH. For Windows MSYS2 user, run shell command like `export PATH=$PATH:$VIVADO_HOME/bin:$VIVADO_HOME/lib/win64.o`. For Linux user just source the Vivado's settings64.sh file located at `VIVADO_HOME`.
+
```
dlmiles, line 81: This is more like, what there reader is looking for :)

It is possible to provide a code snippet of a minimal SpinalSim testbench which can be used as an installation and setup check ?  It helps the reader to know if they run this program and they have everything setup, it should work.  Then in the future with feedback we can address any common error messages seen in the setup and how to fix.  Getting the environment working is probably why the reader is looking at the documentation here.

dlmiles (commented):
Feedback

dlmiles:
Hopefully this will resolve #201  ?

Readon:
> Hopefully this will resolve #201 ?
> 
> Another interesting topic would be how to use a Xilinx primitive example. Maybe that is slightly out of scope for this issue/PR. But if there was a 10 line BlackBox snippet (for a common primitive, IBUF?) included in any SpinalSim example that would be really useful for the reader to see.

please read  the other part of this page. i have borrowed some from them.😁

source/SpinalHDL/Simulation/simulator_specifics.rst
Readon: I don't think this page is a good place to write this details. May be [this location](https://spinalhdl.github.io/SpinalDoc-RTD/master/SpinalHDL/Simulation/bootstraps.html#configuration) is better.

dlmiles: It is the specific detail needed to make XSim work that would be captured here.
I don't think that is best put in the generic simulation information, but maybe a separate page per simulator is a better format.
Readers using XSim want to copy-and-paste something and know it should work as-is.  A kind of installation and setup check, something they can then communicate about if there is an error/issue they can not resolve.

dlmiles:
> please read the other part of this page. i have borrowed some from them.😁

Understand there is a little copy-and-paste going on here.

For example from the VCS/Synopsis section.  I do not know the licensing situation for Synopsis myself (it is not a product I have come into contact with) so I can't comment on that to improve that.
But I do understand more about the Xilinx licensing terms, so this is why I refine the text concerning Xilinx and XSim to provide a better perception of the licensing situation to the reader.  I am not aware of any circumstances for free licensing of Synopsis simulation products.

source/SpinalHDL/Simulation/simulator_specifics.rst
Readon: So that would be more work on this. I just keep the consistency of other parts.

andreasWallner:
I'd be for weakening the language on the license note: XSim is free to use in a wide range of scenarios. Apart from that I'd be happy to approve this - it's ok as it is, even though we could add more like @dlmiles suggested.

source/SpinalHDL/Simulation/simulator_specifics.rst
Readon: > It is the specific detail needed to make XSim work that would be captured here. I don't think that is best put in the generic simulation information, but maybe a separate page per simulator is a better format. Readers using XSim want to copy-and-paste something and know it should work as-is. A kind of installation and setup check, something they can then communicate about if there is an error/issue they can not resolve.

You're right, a simulation tutorial for each simulation engine would be a good choice.

PR #232, opened 2023-12-01
fix index.rst Stream example

typingArtist:
The code seems to be copied from the connectFrom (<<) method which is the inverse method of >> that is described here. Exchanging this for that makes it more logical, I think.


Dolu1990:
Thanks ^^

PR #233, opened 2023-12-17
add chinese version link. Should be fixed after officially merged.

Readon:
This is a thanks to the weblate's hosting of continuous localization.

Now the translated repos is still under SpinalHDL-CN, I still have to find a way to switch the repos link first to have all modifications merge back to upstream.


Dolu1990:
All good :D ?

Readon:
> All good :D ?

Yes，it's good now.

PR #235, opened 2024-01-22
clock_domain: warn about clock wire in SlowArea

louiecaulfield:
Adds a warning about clock usage in SlowArea (see SpinalHDL/SpinalHDL#1287)


Dolu1990:
Thanks !

PR #238, opened 2024-01-28
An initial version of Chinese translation.

Readon:
With the help of weblate we translate the current version of mainstream English version to Chinese.
And merges this into an automated flow.


Dolu1990 (commented):
Look good to me ^^ Thanks !

Maybe one thing to add would be in the readme to document how to generate the Chinese doc.

```
   make -e SPHINXOPTS="-D language='zh_CN'" html # for html in Chinese
```

(i tried to push it, but i do not have the rights)

Dolu1990:
Thanks :D

PR #249, opened 2024-03-30
Add Plru Docs

MahirAbbas:


Dolu1990:
Hi ^^

I would say, the PLRU thing should be documented independently of the Plugin / pipeline framework.

```scala
  val io = new Bundle{
    val context = new Bundle{
      //user -> plru, specify the current state
      val state = Plru.State(entries) 
      //user -> plru, allow to specify prefered entries to remove. each bit set mean : "i would prefer that way to not to be selected by PLRU"
      val valids = withEntriesValid generate Bits(entries bits) 
    }
    val evict = new Bundle{
      //PLRU -> user, Tells you the least recently used entry for the given context provided above
      val id =  UInt(log2Up(entries) bits)
    }
    val update = new Bundle{
      // user -> PLRU specify which entry the user want to mark as most recently used
      val id = UInt(log2Up(entries) bits)
     // PLRU -> user specfy what should then be the new value of the PLRU status 
      val state = Plru.State(entries)
    }
  }
```

MahirAbbas:
Yeah i just copied those over for notes, i was planning to update it so that it's independent of anything else

MahirAbbas:
I haven't fully got round to using it so I won't be able to document it all now. Once I understand it's usage better I should be able to add more docs.

Dolu1990:
That PLRU utils doesn't contain any state, only combinatorial behaviour, usefull in the control logic of a cache to figure out what way to evict, and also logic to update the PLRU with the last way usage

MahirAbbas:
@Dolu1990 is there anything else you want to add or would you be happy with these docs?

source/SpinalHDL/Libraries/Misc/Plru.rst
```diff
@@ -0,0 +1,68 @@
...
+Example usage in a cache 
+
+.. code-block:: scala
+
+   val plru = new Area {
+      // Define a Mem, to track the state of each set
+      val ram = Mem.fill(nSets)(Plru.State(wayCount))
+      val write = ram.writePort
+      val fromLoad, fromStore = cloneOf(write)
+      write.valid := fromLoad.valid || fromStore.valid
```
Dolu1990, line 49: I would say, this is when you have two pipeline (one load and one store) which want to update the PLRU state.
For the doc itself, we shouldn't have that complexity.

There is two example of PLRU usage you may haven't seen : 
- https://github.com/SpinalHDL/VexiiRiscv/blob/b2121967ab0848bf44cfbb8baa3321345f2752bf/src/main/scala/vexiiriscv/fetch/FetchL1Plugin.scala
- https://github.com/SpinalHDL/VexiiRiscv/blob/b2121967ab0848bf44cfbb8baa3321345f2752bf/src/main/scala/vexiiriscv/execute/lsu/LsuL1Plugin.scala

Dolu1990:
Thanks ^^

PR #250, opened 2024-03-30
fix the error caused by wrong format of newly added stage function description.

Readon:
Caught on problem caused by #248. ^^


Dolu1990:
Hooo nice, thanks :D

PR #252, opened 2024-04-18
Add doc for `History(that, range)`

mrcmry:
Based on my limited understanding of [The UART workshop solution](https://github.com/SpinalHDL/SpinalWorkshop/blob/9fd30eb3071b6849ce676fb3086cc2fc22ff731e/src/main/scala/workshop/uart/assets/UartCtrlRx.solution#L34) and [spinalhdl lib code](https://github.com/SpinalHDL/SpinalHDL/blob/768f0960f581c3160c4ef14ebb03928f1f9e2db5/lib/src/main/scala/spinal/lib/Utils.scala#L1218), I updated the documentation. 

Please check that the description is valid, I'm very new to all of that.


Dolu1990:
Thanks :D

PR #255, opened 2024-05-13
Update bits.rst in Bit extraction

uxuh:
x(x: UInt) to x(y: UInt) according to Description "Variable bit access of y-th bit"


Dolu1990:
Thanks :D

PR #257, opened 2024-05-29
whenBuilder document

jijingg:


Dolu1990:
Thanks :)

Readon:
Here is that possible to add foreach in switch example too? I explored the switch usage explanation, and did not find.
So it might be hard to understand the "foreach in switch" statement.

PR #260, opened 2024-06-26
Correct supported slave factory

mrcmry:
Based on implementations present in directory /lib/src/main/scala/spinal/lib/bus of spinalHDL v1.10.2 .


Dolu1990:
Thanks ^^

PR #262, opened 2024-07-20
Update Int.rst about Bit extraction

uxuh:
x(x: UInt) to x(y: UInt) according to Description "Variable bit access of y-th bit"


Readon:
Good catch. Thanks. ^^

PR #277, opened 2025-04-30
Add extra blackbox description and fix typo and presentation

mrcmry:
I added VexiiRiscv reference in Fiber because it's interesting as a code example.

I did not add it to the misc.pipeline doc, because it's probably too complicate as a pipeline example (Vexii redefine a it own class in top of le lib).


Dolu1990:
Nice :D
Thanks

PR #279, opened 2025-08-06
Add additional requirements for compiling verilator using MinGW

vincentbaeten:
Adding some additional requirements that are needed for compiling the latest version of verilator using MinGW on windows.


mrcmry:
Any new about this PR ?

It seems that spinal do not support the latest verilator (see https://github.com/SpinalHDL/SpinalHDL/issues/1919).

Perhaps we should mention this in the doc as well.

PR #283, opened 2025-12-02
docs: Add documentation for partially assigned register design error

### LINKED ISSUE #127: Document design error: Partially assigned register

jooyeon111:
Add documentation for "Partially assigned register" design error.

Fixes #127

### DIFF
--- a/source/SpinalHDL/Design errors/partially_assigned_register.rst
+++ b/source/SpinalHDL/Design errors/partially_assigned_register.rst
@@ -1,0 +1,118 @@
+Partially assigned register
+============================
+
+Introduction
+------------
+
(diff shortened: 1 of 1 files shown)


source/SpinalHDL/Design errors/partially_assigned_register.rst
```diff
@@ -0,0 +1,118 @@
+Partially assigned register
+============================
+
+Introduction
+------------
+
+SpinalHDL will check that when a register is conditionally assigned, all bits of the register must be assigned in every execution path. If some bits of a register are left unassigned in certain conditions, it can lead to unintended hardware behavior such as latches or unpredictable values.
```
Dolu1990: > all bits of the register must be assigned in every execution path

I think it can be confusing, as some could understand it as  :

when X
  need to assign all bits of reg
else
  need to assign all bits of reg
  
  Instead maybe it can be : 
  "all bits of the register must at least be assigned once"   ?

PR #286, opened 2026-02-07
Spelling and fix code example to fit in html box

mrcmry:
If the code do not fit in its box in html, the reader must scroll with the box bar and it disturb the comprehension.

So this PR fix this and also correct some spelling.

Also, logic.rst was refactored to avoid the sphinx warning. As it is included nowhere, there is still a warning.

I also added some UFix/SFix function docs.


Dolu1990:
thanks ^^

PR #299, opened 2026-04-04
Update sphinx package in requirements.txt

mrcmry:
This PR update the python packages and CI yml to the latest version what works on ubuntu 24.04

I  updated the packages to work on 3.12.3, the version of ubuntu 24.04.

I fixed some chinese warnings due to `` not being detected due to missing spaces. Note that I have not knowledge of Chinese or any ideogram languages. But it's probably better like this that without. Perhaps the space before the special parenthesis used could be improved. At least it fix most of the warnings. I left one because it seems to be a multi line problem and I don't want to mess with the meaning.

I pined the ubuntu `runs-on` to 24.04 so it will not fail when a new version is available.

I upgraded `actions` to the latest major (in some place it was using a main or dev branch, I looked at the timeline checked that the released version was older that the blame of the yml file line).

So this will hopefully fix the warnings in Dependabot and probably generate a better doc due to the upgraded libs.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -61,7 +61,7 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v6
       with:
         fetch-depth: 0   ## all branches all tags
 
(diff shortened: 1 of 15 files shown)


mrcmry:
for the record, I force pushed this branch to `dev` to test that the deployment work properly.

Before doing this, I verified that `dev` had no meaningful extra commit compared to master. There where none (it was really outdated).

It seems it's not deploying correctly. I will work on this later.

mrcmry:
This has been tested on dev and is now in #312 for integration in master

PR #305, opened 2026-05-26
Update Symbol to SymbolName in regIf.rst

facebreeze:
The implicit argument of **spinal.lib.bus.regif.RegInst#field** is **SymbolName** instead of **Symbol**


mrcmry (approved):
Thanks !

PR #306, opened 2026-05-27
Update stream.rst to clarify fragmentLock usage

facebreeze:
fragmentLock can be used with **Stream[Fragment[T]]** instead of **Stream[Flow[T]]**


mrcmry (commented):
Thanks !

